Issue #15, opened 2026-09-11
No physical-device smoke has ever been run on this project


RogueKostas:
## Why
Every milestone in `docs/REVIVAL_PLAN.md` from R0 onward records native behaviour as explicitly deferred: no Android emulator, no iOS simulator, no attached device on the build host. The SDK 54→57 upgrade is verified for typecheck, tests and web export only, and F6 makes native startup itself an open question.

The deferred checks named across the plan: Android edge-to-edge tab-bar insets, native stack and modal transitions, drag interactions, dark-theme system-bar contrast, splash appearance, keyboard behaviour, gesture and animation behaviour, persistence across restart, and timer suspend/resume.

## What is needed from the director
A device or an emulator, and the repository linked to the correct **personal** Expo account — `eas.json` and the bundle identifier `com.roguekostas.zenroutine` are already checked in, and the HyperKostas/Hypersonic identity must not be used.

Until then every native claim in this repository stays labelled unverified, and no distributable build or release claim may be made.

Issue #5, opened 2026-09-11
F3: dangling activityTypeId throws unconditionally, locking the user out


RogueKostas:
## Why
`src/store/persistence.ts:342-352` throws on a dangling `activityTypeId` with no `version < CURRENT_SCHEMA_VERSION` gate — unlike the `goalId` linkage check (`persistence.ts:367-388`) and the `routineBlockId` linkage check (`persistence.ts:401-413`), which repair legacy data by clearing the dangling id and only throw at current-version steady state.

The pre-PR store did no structural validation at all (blind `persistedState as AppState`), so any inconsistency already on a real device — a partial write, a hand-edited or restored backup, any path that ever detached an activity type — turns a repairable migration into the same destructive-reset lockout as F1.

Found by independent adversarial review before PR #2 was merged. Severity P2, CONFIRMED.

## Deliver
1. Make the `activityTypeId` check consistent with its two siblings: repair for legacy versions, throw only at current-version steady state.
2. A test for each of the three linkage checks proving legacy data is repaired and current-version corruption is still rejected.

## Gates
- `npm ci` from the tracked lockfile first — a fresh worktree has no dependencies.
- `npm run typecheck` — exit 0, no new `any`, `@ts-ignore` or `@ts-nocheck`.
- `npm test` — exit 0, and the suite count must not go down.
- `npm run build:web` — exit 0.
- Add a **negative control** for every new test: make it fail deliberately once, record that it went red, then restore it. A check that cannot go red is not a check.

## Do not
- Do not push, open a PR, merge, or run `gh auth`, `git stash`, `git reset --hard`, `npm audit fix`, `expo prebuild`, `eas build` or any Render command.
- Do not touch files outside the scope above — other lanes are working in this repo.
- Do not regenerate `package-lock.json`; it is tracked, reviewed truth.
- Corrections to this brief are welcome and expected. If the framing here is wrong, say so in your report.

## Report must include
Files or behaviours examined, command output as evidence, actionable findings, and a **Not done / unverified** section.

## Done means
The three linkage checks behave identically with respect to version, and a legacy store with a dangling activity type hydrates with the reference cleared.

RogueKostas:
**Scope widened by the orchestrator, after the non-author review of PR #17.**

This issue was originally only about making the dangling `activityTypeId` check version-consistent
with its two siblings. The review of PR #17 found that the linkage checks are a **second live route
to the same permanent brick that #3 was filed for**, and that is now part of this issue.

[...]

Issue #4, opened 2026-09-11
F2: legacy migration silently discards the duration of a second open timer


RogueKostas:
## Why
`src/store/persistence.ts:419-429`: when migrating schema < 4 data containing more than one tracking entry with `endTime === undefined`, the non-selected open entries are force-closed with `endTime = startTime`, discarding their real elapsed duration with no warning to the user.

This is currently codified as intended behaviour — `tests/store/persistence.test.ts:93-114` asserts `openB.endTime === openB.startTime`.

Found by independent adversarial review before PR #2 was merged. Severity P1, CONFIRMED.

## Deliver
1. Decide the correct behaviour and state it in the code: the most defensible option is to close the entry at the last moment there is evidence for (for example the entry's own last-updated timestamp, or the start of the next entry for the same activity), never at `startTime`, and to record that the entry was repaired.
2. Surface repaired entries to the user rather than mutating history silently.
3. Update `tests/store/persistence.test.ts` so it asserts the new contract, and add a test that the repair is reported.

## Gates
- `npm ci` from the tracked lockfile first — a fresh worktree has no dependencies.
- `npm run typecheck` — exit 0, no new `any`, `@ts-ignore` or `@ts-nocheck`.
- `npm test` — exit 0, and the suite count must not go down.
- `npm run build:web` — exit 0.
- Add a **negative control** for every new test: make it fail deliberately once, record that it went red, then restore it. A check that cannot go red is not a check.

## Do not
- Do not push, open a PR, merge, or run `gh auth`, `git stash`, `git reset --hard`, `npm audit fix`, `expo prebuild`, `eas build` or any Render command.
- Do not touch files outside the scope above — other lanes are working in this repo.
- Do not regenerate `package-lock.json`; it is tracked, reviewed truth.
- Corrections to this brief are welcome and expected. If the framing here is wrong, say so in your report.

## Report must include
Files or behaviours examined, command output as evidence, actionable findings, and a **Not done / unverified** section.

## Done means
No migration path can silently zero a duration the user actually worked, and the test suite asserts that rather than the old behaviour.

PR #17, opened 2026-09-11
fix(store): stop a backwards clock from bricking hydration (#3)

### LINKED ISSUE #3: F1: stopTracking can persist endTime < startTime and permanently brick hydration
## Why
`src/store/useAppStore.ts:766` (`stopTracking`) writes a tracking entry without validating `endTime >= startTime`, unlike `addCompletedEntry` and `updateTrackingEntry`, which both call `trackingEntryIsValid`.

If the device clock moves backwards between start and stop — NTP resync, manual clock correction, timezone/clock skew after reconnecting — the resulting entry has `endTime < startTime` and is persisted to AsyncStorage.

On the next launch, `merge()` (`useAppStore.ts:1015-1021`) always re-runs `migratePersistedState(persistedState, CURRENT_SCHEMA_VERSION)`. Because every save stamps the current version (`useAppStore.ts:263,1011`), this is always the strict path, so `parseTrackingEntry` (`src/store/persistence.ts:286-289`) throws `Invalid tracking entry: endTime is before startTime` instead of repairing.

Hydration then fails permanently. "Try again" re-hits the same record. The only recovery the UI offers (`App.tsx:79-123`, "Reset local data") wipes every goal, routine and tracking record on the device.

Found by independent adversarial review before PR #2 was merged. Severity P1, CONFIRMED by code trace.

## Deliver
1. Validate in `stopTracking` on the same terms as `addCompletedEntry`/`updateTrackingEntry`. Decide and document what happens to a backwards-clock session — clamping to a zero-length entry, discarding it with a user-visible message, or storing it flagged — rather than silently persisting an invalid one.
2. Make hydration survivable: a single unparseable tracking entry must not be able to block the whole store. Quarantine or drop the offending record, keep the rest of the user's data, and surface what was dropped.
3. Regression tests for both: a backwards-clock stop, and hydration of a store that already contains one corrupt entry (the on-disk state a user could already have).

## Gates

## Do not

## Report must include

## Done means
A device that has already persisted a corrupt entry can launch, keep its goals and routine, and be told what was dropped — proven by a test that fails without the fix.

RogueKostas:
Closes #3.

Lane `f1-hydration`, headless `claude -p` in its own worktree, branched from `261c592`.
Two dispatches: the first produced the implementation but every gate came back `MISSING` because
the worktree carried no allow list; the second ran the gates and committed.

## What it does

1. **`stopTracking` clamps instead of persisting an invalid entry.** A backwards clock now writes
   `endTime = startTime` (zero-length) rather than `endTime < startTime`. The reasoning is in a
   comment at the site: discarding is the one irreversible outcome on an offline-first device, a
   zero-length entry contributes 0 minutes so a bad clock can neither invent nor erase goal
   progress, and it survives the strict migration path.
2. **Hydration survives a corrupt record.** `migratePersistedState` takes an optional quarantine
   sink; an unparseable tracking entry is set aside and skipped instead of aborting the load.
   Activity types, goals, routines, and `trackingEntries` not being an array all stay strict.
   **Backup import stays strict** by omitting the sink — an import is a deliberate act on a
   re-choosable file and should fail loudly.
3. **Nothing is lost silently.** Quarantined records are copied verbatim to a separate
   `zenroutine-quarantine` key *before* anything can rewrite the store blob, and `App.tsx` shows a
   dismissible notice.

## Gates — run by the lane, re-run by the orchestrator on the merged tree

| Gate | Lane |
|---|---|
| `npm ci` from tracked lock | exit 0, lockfile unchanged after |
| `npm test` | exit 0, **72 passed** (62 at `origin/main`, +10) |
| `git diff --name-status origin/main...HEAD` | 7 `M`, no `D` |

Eight negative controls, each taken RED and restored to GREEN. The one worth naming: disabling
quarantine capture reproduced issue #3 verbatim — `status: 'error'`,
`"Invalid tracking entry: endTime is before startTime"`.

## The lane's corrections to its own earlier work

Re-dispatched with the resume preamble, it found a **real bug in its first attempt**: the pointer
rule cleared `currentTrackingEntryId` by searching the quarantine list for a matching `id`, so a
record whose *own* `id` is the unreadable field gets quarantined as `id: null`, the search misses
it, and hydration still throws `Invalid currentTrackingEntryId` — still bricked, by the exact
mechanism being fixed. The rule now asks whether the pointer still resolves, gated on something
actually having been quarantined so a dangling pointer with no bad record to blame stays loud.
Negative control 3 is that bug going red.

It also corrected a comment that claimed the clamp made the action and persistence layers agree —
they do not, since `trackingEntryIsValid` requires `endTime > startTime`. The clamp is still right;
the asymmetry is deliberate and now documented as such. And the quarantine side-car appends rather
than overwriting, so a second event cannot destroy the first generation's records.

The brief's line numbers and its reading of the merge path were verified against `origin/main` and
were all correct.

## NOT OBSERVED

- **No physical device run.** Verified on the vitest AsyncStorage mock and the web export only.
  #15 remains open and this does not close it.
- **The `App.tsx` quarantine banner has zero automated coverage** — never rendered in any test.
  Layout, dismiss button and `accessibilityLiveRegion` are unverified anywhere. Weakest part of
  the change.
- **A real backwards clock was never exercised** — tests move a fake timer.
- **The side-car is write-only.** Records are retained under `zenroutine-quarantine` but nothing
  reads them back; the banner says records were set aside and cannot show them. Worth a product
  decision.
- **Duplicate tracking-entry ids still abort hydration** — a second route to the same brick, not
  covered here.

Issues #4 and #5 (the legacy open-timer block and the dangling `activityTypeId` check) were left
untouched, as briefed.


RogueKostas:
## Orchestrator review — non-author, before merge

**Gate run by me on the merged tree**, at `370fcac`, verified as the merged tree
(`git rev-list --count HEAD..origin/main` = 0), on the host with Node v22.23.2:

| Gate | Result |
|---|---|
| `npm ci` | exit 0 |
| `npm run typecheck` | exit 0 |
| `npm test` | exit 0 — 7 files, **72 passed** (62 at `261c592`) |
| `npm run build:web` | exit 0, `Exported: dist` |

Gated push: 0 deletions before and after syncing `origin/main`, tracked tree clean, remote ref read
back and matching.

Then an independent non-author review of the diff. **Merging**, because this fixes a real
unrecoverable brick and introduces no regression — every remaining path to that brick predates this
branch. Five findings, none of which argue for holding it, all filed:

### Confirmed clean
The strictness claims are true as written: `activityTypes`, `goals` and `routines` all parse with no
try/catch and throw straight through the quarantine option, and backup import really does stay
strict (`decodeBackup` calls `migratePersistedState` with no options). The clamp is well reasoned and
destroys nothing a stopped clock had not already destroyed. The tests assert behaviour — state shape,
quarantine contents, thrown messages, storage contents — not implementation, and would fail if the
clamp or the strict/lenient split were reverted.

### P1 — a linkage failure on a well-formed entry still bricks hydration → widened into #5
`src/store/persistence.ts:485-508,513-533`. `hasInvalidEntryGoal` / `hasInvalidEntryRoutineBlock` are
computed over all entries and throw unconditionally on the current-version path, entirely outside the
quarantine-catching `flatMap`. An entry that parses perfectly but whose `goalId` no longer matches any
goal — a goal record lost to a torn write, a bad manual edit, a buggy restore — reproduces issue #3's
failure mode for a different corruption shape.

This is **not a regression and not a gap in this lane's work**: the brief explicitly reserved the
linkage checks for #4 and #5, and the lane flagged it in its own report. #5's scope is now widened to
quarantine linkage failures rather than only making the three checks version-consistent.

### P1 — the side-car can rotate out a genuine generation → #18
The branch's own test pins that the store blob is left byte-identical after a merge-only quarantine,
so the bad record stays on disk and re-quarantines on every cold start until some unrelated write
rewrites the blob. With `MAX_QUARANTINE_GENERATIONS = 20` and `.slice(-20)`, twenty cold starts of
the same uncleaned record silently drop the oldest generation — which may be a genuinely different
earlier event. The comment at `persistence.ts:64-66` calls this "realistically unreachable"; the
test at `hydrationAndBackup.test.ts` asserting the blob is unchanged is what makes it reachable.

### P1 plausible — a write-back race on pre-v4 blobs → #19
Zustand's `persist` calls `migrate()` only on a version mismatch and, when it does, persists the
migrated state before `rehydrate()` resolves — i.e. before the side-car append runs in
`initializeAppStore`. On a legacy blob, the record could be stripped from the app blob and then lost
from the side-car too if that write fails (it only `console.warn`s) or the process dies in the gap.
Not traced against the installed `zustand@5.0.11`, so it is filed as measure-and-report, not as a fix.

### P2 — the pointer-clear gate does not check the blame relationship → #20
`persistence.ts:536-554` requires `quarantine.length > 0` but never that the dangling id belongs to
what was quarantined. A stale `currentTrackingEntryId` that never referred to anything real is
silently nulled whenever *something else* was quarantined, so genuinely unrelated corruption is
swallowed — contradicting the comment two lines above it, which says a dangling pointer with no bad
record to blame should stay loud.

### P2 — the notice's dismissal is not scoped to an event → #21
`App.tsx:56-60,143-145`. `quarantineDismissed` is a bare `useState(false)` that is never reset, so a
later unrelated quarantine in the same mounted instance is never reported.

Issue #6, opened 2026-09-11
F4: forecast confidence is scoped to activity type, not to the goal


RogueKostas:
## Why
`src/core/engine/prediction.ts:66-75` (`getForecastEvidence`) filters tracking history by `activityTypeId` only. It never checks `entry.goalId`. The resulting `confidenceLevel` and `confidenceReason` are then shown per goal in `GoalsScreen.tsx:289-292`.

Failure scenario: two active goals share the activity type "study". Every tracking entry in history belongs to goal A. Goal B, which the user has never worked on, still counts all of A's distinct evidence days and is shown a high-confidence completion date.

The branch's own test (`tests/core/prediction.test.ts:102-134`) does not expose this only because the competitor goal's allocation is zero, which short-circuits before evidence is counted. Give it any allocation and the bug surfaces.

Found by independent adversarial review before PR #2 was merged. Severity P1, CONFIRMED.

## Deliver
1. Scope evidence to the goal. Decide explicitly how entries with no `goalId` are treated — they are ambiguous, and the decision needs to be written down next to the code, not inferred.
2. A test with two active co-allocated goals of the same activity type and different histories, asserting their confidence differs.

## Gates
- `npm ci` from the tracked lockfile first — a fresh worktree has no dependencies.
- `npm run typecheck` — exit 0, no new `any`, `@ts-ignore` or `@ts-nocheck`.
- `npm test` — exit 0, and the suite count must not go down.
- `npm run build:web` — exit 0.
- Add a **negative control** for every new test: make it fail deliberately once, record that it went red, then restore it. A check that cannot go red is not a check.

## Do not
- Do not push, open a PR, merge, or run `gh auth`, `git stash`, `git reset --hard`, `npm audit fix`, `expo prebuild`, `eas build` or any Render command.
- Do not touch files outside the scope above — other lanes are working in this repo.
- Do not regenerate `package-lock.json`; it is tracked, reviewed truth.
- Corrections to this brief are welcome and expected. If the framing here is wrong, say so in your report.

## Report must include
Files or behaviours examined, command output as evidence, actionable findings, and a **Not done / unverified** section.

## Done means
A goal with no tracking of its own never reports high confidence because a sibling goal has history.

PR #22, opened 2026-09-11
fix(prediction): scope forecast confidence to the activity type that changed (#7)

### LINKED ISSUE #7: F5: editing any routine block collapses every goal's forecast confidence
## Why
`src/core/engine/prediction.ts:71` compares each tracking entry against `routine.updatedAt` — the whole routine's timestamp — rather than anything scoped to the activity type or block that changed. `addRoutineBlock`, `updateRoutineBlock` and `deleteRoutineBlock` (`src/store/useAppStore.ts:598,628,~648`) all stamp the entire routine's `updatedAt` on any block mutation.

Failure scenario: a user has 60 distinct tracked days of "Reading" against a long-standing dedicated block, correctly shown as high confidence. They nudge an unrelated "Exercise" block by five minutes. Every one of those 60 days is now older than `routine.updatedAt`, evidence collapses to zero, and the app reports "low confidence — no completed tracking days since this routine changed".

This fires constantly for any routine with more than one activity type, which is the normal case.

## Deliver
1. Invalidate evidence at the granularity that actually changed. Per-block or per-activity-type change timestamps are the obvious route; whichever is chosen, only capacity relevant to a goal should reset that goal's evidence.
2. Tests: editing an unrelated block preserves confidence; changing the capacity that feeds a goal does reset it.

## Gates

## Do not

## Report must include

## Done means
Editing a block of a different activity type leaves the affected goal's confidence and evidence-day count unchanged.

RogueKostas:
Closes #7.

Lane `f5-confidence-scope`, headless `claude -p` in its own worktree. Two dispatches: the first
produced the implementation with every gate `MISSING` (no allow list in the worktree); the second
ran the gates and committed.

## What it does

Forecast evidence is now counted from `Routine.capacityChangedAt[activityTypeId]` instead of
`routine.updatedAt`, so editing a block of one activity type no longer wipes the evidence window for
goals of every other activity type.

- `src/core/types/Routine.ts` — new **optional** `capacityChangedAt?: Record<ActivityTypeId, string>`.
- `src/core/engine/prediction.ts` — new `getCapacityChangedAt(routine, activityTypeId)`; the three
  `getForecastEvidence` call sites use it. **`getForecastEvidence`'s goal-vs-activity filtering is
  untouched** apart from renaming its fourth parameter, so #6 can follow cleanly in the same file.
- `src/store/useAppStore.ts` — `withCapacityChangedAt`; the block mutations stamp only the activity
  types they actually change. `updateRoutineBlock` stamps both the old and the new type, so a
  re-type or re-link also invalidates the side that lost capacity.
- `src/store/persistence.ts` — parse the new field.
- `docs/REVIVAL_PLAN.md` — the R3 correction note.

**Why activity type rather than per block:** goals sharing an activity type draw on the same
unlinked pool and inherit each other's reallocations in `predictActivityGoals`, so a change to any of
that type's capacity genuinely can move all of their dates. Across activity types `predictAllGoals`
keeps the groups independent — which is exactly where the old blast radius was wrong. Per-block would
have been finer than the model is.

## Persisted shape — it changes

`Routine` gains one optional field. **Schema version stays 4** and no data is rewritten on upgrade.

- **Existing data:** routines with no map fall back to `Routine.updatedAt` — precisely the window
  they are already read under, so no existing user's confidence changes at the moment of upgrade.
- **First block edit after upgrade:** every activity type already in the routine is seeded with the
  routine's *previous* `updatedAt`, then the changed ones are stamped forward. Without this seeding
  the fix would not work at all on existing data — untouched types would fall through to the freshly
  bumped `updatedAt` and collapse anyway.
- **Malformed entries** are dropped on read and take the same fallback, rather than failing
  hydration. Deliberately more forgiving than the rest of `persistence.ts`, because this field is
  app-derived, never user-authored, and always has a defined fallback. Flagged in case you want it
  strict.

| Gate | Lane |
|---|---|
| `npm ci` from tracked lock | exit 0, 584 packages, never regenerated |
| `npm run typecheck` | exit 0, no diagnostics, no new `any` / `@ts-ignore` / `@ts-nocheck` |
| `npm test` | exit 0, **69 passed** (62 + 7) |
| `git diff --name-status origin/main...HEAD` | 7 `M`, no `D` |

Seven negative controls, RED then GREEN. Two are worth naming: making `getCapacityChangedAt` ignore
the map reproduced #7 exactly — `high`/14 days → `low`/0 from an unrelated edit; and dropping the
seeding loop reproduced the existing-user case, where the untouched activity type gets no map entry
on the first post-upgrade edit and collapses anyway. That second failure mode is easy to miss by
inspection and is now pinned.

## The lane's corrections to my brief — both accepted

1. **My brief forbade touching `src/store/persistence.ts`. That was wrong.** `parseRoutine` builds an
   explicit object literal and `merge` runs `migratePersistedState` on every hydration, so any new
   field on `Routine` is silently dropped on every reload. A persisted timestamp is not possible
   without a parser change. The lane kept it to one function plus one line, +27, away from the
   hydration path. Posted on #7.
2. **My brief named three block mutations; there are four.** `copyDayBlocks`
   (`src/store/useAppStore.ts:712`) had exactly the same blast radius and would have been a live
   hole in the fix.

It also found that `duplicateRoutine` spreading the source routine carries the map into the copy —
which turns out to be *required*, since dropping it would send the duplicate to the
`updatedAt = now` fallback and reintroduce the bug. No change needed, but it was unverified before.

## NOT OBSERVED

- **Nothing was run on a device or in a browser.** jsdom/vitest plus a successful web bundle only.
- **The migration has never met real user data** — only synthetic persisted payloads.
- **No screen reads `capacityChangedAt` yet**, so this fix is currently visible only through the
  engine. The `GoalsScreen` confidence copy still reads "since this routine changed" and has not
  been seen rendering against migrated data.
- **`updateRoutine` (rename, activate) is a known remainder.** It bumps `updatedAt` without stamping
  the map, so a pre-migration routine renamed *before* any block edit still collapses through the
  fallback. Recorded in `docs/REVIVAL_PLAN.md` and deliberately left, to keep this diff narrow
  enough for #6 to follow in the same file.
- The 7/14 distinct-day thresholds remain the unvalidated prototype assumption R3 recorded. This
  lane changed *when the clock starts*, not what counts as enough.


RogueKostas:
## Orchestrator review — non-author, before merge

**Gate run by me on the merged tree**, at `f9081c3`, verified as the merged tree
(`git rev-list --count HEAD..origin/main` = 0). This one mattered more than usual: `persistence.ts`,
`useAppStore.ts` and `useAppStore.test.ts` all **auto-merged** against the F1 work that landed
minutes earlier in PR #17, so the combination had never been compiled or run anywhere before this.

| Gate | Result |
|---|---|
| `npm ci` | exit 0 |
| `npm run typecheck` | exit 0 |
| `npm test` | exit 0 — **79 passed** (72 after PR #17, +7) |

Gated push: 0 deletions before and after sync, remote ref read back and matching.

**Merging**, and filing three follow-ups. The reasoning on the first one matters, so it is stated
plainly rather than buried.

### The headline fix works. The dominant real-world trigger survives it. → filed as p0

`src/store/useAppStore.ts:689-725`: `updateRoutineBlock` builds `updatedBlock = { ...block, ...data }`
and then calls `withCapacityChangedAt(...)` **unconditionally** — there is no check that anything
actually changed. And `src/components/routine/BlockEditor.tsx:113-121` has no dirty check: Save
always calls `updateRoutineBlock` with the current on-screen values.

So a user who opens an existing Fitness block, changes nothing (or only relinks its goal), and taps
Save still stamps `capacityChangedAt[Fitness] = now`. Every Fitness goal drops from "high confidence,
20 days" to "low confidence — no completed tracking days since this routine changed", with no
capacity change at all.

This is **not a regression** — before this branch, that same tap collapsed confidence for *every*
activity type, not just one — so the branch is a strict improvement and the gate is green, which is
why it lands. But the issue #7 headline ("editing a block no longer collapses unrelated goals") is
only two thirds true until this is fixed, and it is reachable on **every** block edit, so it is the
next lane rather than a backlog item.

### A second writer of `updatedAt` the remainder note does not mention → filed
`src/store/useAppStore.ts:610-620`. `setActiveRoutine` maps over **all** routines and stamps
`updatedAt: now` on each, whether or not that routine's `isActive` changed. Any routine whose map is
not yet fully seeded falls back to `updatedAt` and collapses — issue #7's failure mode with no block
touched. `docs/REVIVAL_PLAN.md:117` names only `updateRoutine` as the remainder; this is a second
instance. Narrow today (the switcher is `__DEV__`-gated, though `RoutineScreen.tsx:166-167` reaches
it in production by deleting the active routine), but the `Routine` type's own "Vacation Mode"
comment says multi-routine switching is intended, so it is a primed landmine rather than a
non-issue.

### Latent, plus the test gap that let the above through → filed
`updateRoutine`'s signature is `Partial<Omit<Routine, 'id' | 'createdAt' | 'updatedAt' | 'blocks'>>`
— every other field this change must protect is either excluded or routed through
`withCapacityChangedAt`, but `capacityChangedAt` is neither, so a future call site could clobber the
whole map. No current call site does. And `duplicateRoutine` carrying the map is correct as traced —
the spread does carry it, nothing mutates it in place — but has **zero** test coverage on either
branch, so a later field-by-field refactor would silently drop it with nothing to catch it.

The store-layer tests exercise `updateRoutineBlock` only with data that genuinely changes, and never
call `setActiveRoutine` or `updateRoutine` in the capacity-timestamp block. That gap is what let the
first two findings through.

### Confirmed clean
The auto-merge with PR #17 is semantically sound: `readCapacityChangedAt`
(`src/store/persistence.ts:370-386`) never throws and touches no shared state, so it cannot interact
with the tracking-entry quarantine path or the two-stage migrate/merge flow. `copyDayBlocks` stamps
both the vacated and the newly-occupied activity types correctly. `withCapacityChangedAt` itself is
correctly ordered — seed, then overwrite — and no sequence of the four block mutations seeds at the
wrong moment or double-stamps; the defects are in what calls it, not in it. The prediction-layer
tests are behaviour-based and would fail on a revert.

Both of the lane's corrections to my brief are accepted and recorded on #7: `persistence.ts` had to
be touched, and there were four block mutations, not the three I named.

PR #25, opened 2026-09-12
fix(prediction): scope forecast confidence to the goal, not the activity (#6)

### LINKED ISSUE #6: F4: forecast confidence is scoped to activity type, not to the goal

### LINKED ISSUE #9: F7: no test isolates per-goal forecast evidence
## Why
`tests/core/prediction.test.ts` has no case constructing two active, co-allocated goals of the same activity type with different tracking histories, and none that edits an unrelated routine block and re-runs the prediction. That gap is what let F4 and F5 through.

The allocation arithmetic tests were independently re-derived during review and are sound — this issue is only about the confidence/evidence feature.

## Deliver
Tests that would have caught F4 and F5, written so they fail against the current implementation. They may land before or after the fixes; if before, mark them appropriately rather than weakening the assertion.

## Gates

## Do not

## Report must include

## Done means
Both tests exist, and each was observed failing before the corresponding fix.

RogueKostas:
Closes #6. Also closes #9 — see below.

Lane `f4-goal-evidence`, headless `claude -p` in its own worktree from `9cca4f2`.

## What it does

`getForecastEvidence` filtered tracking history by `activityTypeId` alone and never looked at
`entry.goalId`, so a goal the user had never worked on inherited a sibling goal's evidence and was
shown a high-confidence completion date. Evidence is now scoped to the goal.

**The rule chosen:** an entry with **no `goalId`** counts as evidence for every goal of that activity
type; an entry linked to a **different** goal never counts.

The defence is in the code comment, and it is sound: unlinked time is precisely what the model puts
in the shared pool and divides among these goals by priority, so observing it *is* observing the pool
the forecast spends. It is also the ordinary path, not an edge case — an unlinked routine block
starts an unlinked entry (`HomeScreen.tsx:68`) and the quick-start picker offers "no goal" outright
(`QuickStart.tsx:60-68`). Discarding it would pin a diligent user at low confidence forever. Time
explicitly attributed to *another* goal, by contrast, is evidence the session went elsewhere — the
opposite of support. That asymmetry is the fix.

Two consequences the lane accepted rather than hid, both stated in the comment: co-allocated goals
with no linked history of their own share one evidence count (honest — nothing distinguishes them
yet), and since deleting a goal clears `goalId` from its entries, a deleted goal's history joins the
unlinked pool.

## The lane's correction to my brief — and it matters

My brief presented the unlinked-entry question as fully open. **It was not.** The lane tested the
strict alternative — unlinked counts for nobody — as a negative control and it turns **six** tests
red: its own new one plus five pre-existing, including *both* of PR #22's `capacityChangedAt` tests
that my brief said must keep passing untouched. Every existing evidence fixture is unlinked, because
`makeTrackingEntry` sets no `goalId`. So "unlinked counts for nobody" and "#22's tests pass
untouched" cannot both hold.

The lane says it would have made the same call on the merits, and flagged that the choice was
constrained rather than free. That is exactly the correction I want a lane to make.

| Gate | Lane |
|---|---|
| `npm ci` from tracked lock | exit 0, never regenerated |
| `npm test` | exit 0, **82 passed** (79 at `origin/main`, +3, no existing test modified) |

Two negative controls, because one mutation cannot redden all three new tests. **Control A**
(revert to activity-only filtering) printed the P1 bug verbatim: `goal-b`, never worked on, returned
`"confidenceLevel": "high"`, `"Based on tracking across 14 distinct days."` **Control B** (unlinked
counts for nobody) is the one described above.

## #9 (F7) is covered by this

`reports different confidence for co-allocated goals with different histories` (14 days vs 7 → high
vs medium) and `never lets a goal inherit a sibling goal's evidence` (0 days → low, through both
`predictAllGoals` and `predictGoalCompletion`). That is precisely the gap #9 was filed for, so #9
closes here.

## NOT OBSERVED

- **No device smoke.** #15 is open and only a device settles it. Verified through the node suite and
  the web export; the changed confidence badge has not been seen rendering for two co-allocated
  goals in any browser.
- **The unlinked decision is a product judgement, not a measurement.** Nobody knows how often real
  users link tracking to a goal. If most tracking is unlinked in practice, co-allocated goals will
  look identical to one another — correct under this rule, but possibly not what D5's tappable
  breakdown should show. This is the thing to overrule, if anything is.
- All evidence is synthetic fixtures; behaviour against a migrated on-device store is unverified.
- The 7/14-day thresholds remain unvalidated prototype assumptions, unchanged here.


RogueKostas:
Orchestrator review, non-author. Gate run by me on the merged tree at `1be6543` (verified merged: 0 commits behind origin/main) — `npm ci` 0, `typecheck` 0, **82 tests**, `build:web` 0. Gated push: 0 deletions before and after sync, remote ref read back and matching.

I read the diff rather than the summary. It is minimal and correct: one predicate, three call sites, no existing test touched. The rule is documented at the function with its reasoning and both of its accepted consequences, which is the standard I want for a judgement call embedded in code.

The lane's correction stands and is the most valuable thing in this PR: the unlinked-entry question was **not** open, because the strict alternative reddens five pre-existing tests including both of PR #22's. My brief said otherwise. Recorded rather than quietly dropped.

Merging. #9 closes with it — the two new tests are exactly the coverage that issue was filed for.

PR #26, opened 2026-09-12
fix(store): stop a no-op block save from collapsing confidence (#23)

### LINKED ISSUE #23: A no-op block save still collapses that activity type's confidence
## Why

Found in the non-author review of PR #22 (F5). **P1, CONFIRMED. This is the dominant real-world
trigger of issue #7, and it survived the fix for it.**

`src/store/useAppStore.ts:689-725`: `updateRoutineBlock` builds
`updatedBlock = { ...block, ...data }` and then calls
`withCapacityChangedAt(routine, [block.activityTypeId, updatedBlock.activityTypeId], now)`
**unconditionally**. Nothing checks whether `updatedBlock` differs from `block`.

`src/components/routine/BlockEditor.tsx:113-121` has no dirty check either: when `isEditing`, Save
always calls `updateRoutineBlock(routineId, block.id, { startMinutes, endMinutes, activityTypeId,
goalId })` with whatever is currently on screen.

Failure scenario: a user opens an existing "Fitness" block to look at it — or only to relink its goal,
leaving the activity type and times alone — and taps Save. `capacityChangedAt['fitness']` is stamped
to `now`. Every Fitness goal drops from "high confidence, tracked across 20 days" to
"low confidence — No completed tracking days since this routine changed". No capacity changed.

PR #22 was still a strict improvement — before it, that same tap collapsed *every* activity type —
which is why it landed. But this path is reachable on **every** block edit, so the #7 headline is not
actually delivered until this is fixed.

## Deliver

1. Stamp only when capacity actually changed. The capacity-relevant fields are the activity type, the
   day, the start and end minutes, and the goal link (a goal link moves time between the dedicated
   and shared pools, so it counts). A change to anything else — or to nothing — must not stamp.
   Write the list of capacity-relevant fields down in the code; the next reader should not have to
   infer it.
2. Consider whether `BlockEditor` should avoid calling the store at all on an unchanged save. Fix the
   store regardless — the store must be correct whatever the UI does — and say in your report whether
   the UI change is also worth making.
3. Tests: a no-op save leaves `capacityChangedAt` untouched; a goal-only relink DOES stamp; a time
   change stamps; an activity-type change stamps both the old and the new type (this already works —
   do not break it).

## Gates

`npm ci` from the tracked lock first (a fresh worktree has none), then `npm run typecheck`,
`npm test`, `npm run build:web`, each in its own process. A negative control per new test: RED
before, GREEN after, both pasted. Suite was **79** at the merge of PR #22; it must not go down.

## Scope

`src/store/useAppStore.ts` (the block mutations only) and `src/components/routine/BlockEditor.tsx`,
plus tests. **Do not touch** `src/core/engine/prediction.ts` or `tests/core/prediction.test.ts` — a
lane is working there on #6. Do not touch the quarantine or hydration paths in
`src/store/persistence.ts`.

## Do not

## Done means

Opening a block and tapping Save without changing anything leaves every goal's confidence and
evidence-day count exactly as it was — proven by a test that fails without the fix.

RogueKostas:
Closes #23.

Lane `no-op-save`, headless `claude -p` in its own worktree from `9cca4f2`. The lane reports no
corrections to the brief's core diagnosis — it held up exactly — and one refinement, flagged below.

## What it does

`updateRoutineBlock` stamped `capacityChangedAt` unconditionally, and `BlockEditor` has no dirty
check, so opening a block and tapping Save with nothing changed collapsed that activity type's
confidence. PR #22 fixed the blast radius across activity types; this fixes the trigger.

- **Nothing changed → nothing written.** Early return: no new `updatedAt`, no stamp, no persist write.
- **Something changed → stamp only if capacity moved**, against a named constant:

```ts
const CAPACITY_RELEVANT_BLOCK_FIELDS: readonly (keyof RoutineBlock)[] = [
  'activityTypeId', 'dayOfWeek', 'startMinutes', 'endMinutes', 'goalId',
];
```

Written out explicitly rather than derived from `RoutineBlock`, with a doc comment giving the reason
each field is on the list, so a field added later has to be opted in rather than silently collapsing
confidence. `id` is deliberately absent. Both sides of an activity-type change still stamp; a
goal-only relink still stamps.

## The UI half — the lane's answer is no, and I agree

`BlockEditor` gets no dirty check. With the store inert on a no-op, the only side effect of a
look-and-Save is a function that returns early; `onSave` is just `handleCloseEditor`
(`RoutineScreen.tsx:221`), so there is nothing else to suppress. A UI dirty check would duplicate the
capacity rule in a second place that can drift from the first. `BlockEditor.tsx` is unchanged.

| Gate | Lane |
|---|---|
| `npm ci` from tracked lock | exit 0, never regenerated |
| `npm run typecheck` | exit 0, no diagnostics |
| `npm test` | exit 0, **83 passed** (79 at `origin/main`, +4) |

Four negative controls. The first is the real one — the no-op save test went RED pre-fix with
`expected 09:01:00 / received 09:02:00`, the stamp firing on a save that changed nothing. The other
three tests guard the fix from *over*-reaching, so they pass pre-fix by construction; their control
is mutation testing, and each mutation was confirmed to redden its test: dropping `'goalId'` from
the list, dropping `'dayOfWeek'`, and stamping only the new activity type instead of both. All
mutations reverted and the restored source diff-verified before commit.

## NOT OBSERVED

- **No device or browser run.** Every claim is proven against the store through vitest. That
  `BlockEditor` sends an unchanged payload on a look-and-Save is read from `BlockEditor.tsx:116-122`,
  not watched in a running app.
- **`dayOfWeek` is a judgement call, not a measurement**, and the lane says so. My brief listed it as
  capacity-relevant and it was kept, but `getWeeklyMinutesForActivityType` sums across the whole week
  — moving a block Monday→Tuesday does not change the minute total, and nothing else in
  `prediction.ts` reads `dayOfWeek`. It stamps anyway because it changes the schedule the evidence
  was gathered against, which is what confidence claims to measure. If a day move should not cost a
  confidence reset, dropping it from the list is a one-line change and control 3 is the test that
  will tell you.


RogueKostas:
Orchestrator review, non-author. Gate run by me on the merged tree at `5dee5f2` (0 commits behind origin/main) — `npm ci` 0, `typecheck` 0, **86 tests**, `build:web` 0. Gated push: 0 deletions before and after sync, remote ref read back and matching.

I read the diff. Two things I checked specifically and am satisfied with: the early return fires before `updatedAt` is touched at all, so a no-op save now writes literally nothing; and `changedFields` compares the merged block against the stored one rather than trusting the payload's shape, so a field that is present but identical does not count as a change.

The named constant with a per-field justification is the right call — it makes a future field an explicit opt-in rather than a silent confidence reset, which is exactly the failure this issue was.

Agreeing with the lane on the UI half: no dirty check in `BlockEditor`. With the store inert, a look-and-Save has no remaining side effect, and a second copy of the capacity rule in the component is a thing that drifts.

On `dayOfWeek` — the lane flagged it as a judgement rather than a measurement, correctly. Weekly capacity sums across the week so a Monday-to-Tuesday move changes no minute total, but it does change the schedule the evidence was gathered against. Keeping it. Noted here so that if a day move turns out to cost a confidence reset users resent, the decision is findable and reversible in one line.

Merging.

Issue #24, opened 2026-09-11
setActiveRoutine and updateRoutine bump updatedAt without stamping the capacity map


RogueKostas:
## Why

Found in the non-author review of PR #22 (F5). **P1 as a code defect, narrow reachability today.**

`src/store/useAppStore.ts:610-620`, unchanged by that PR:

```js
routines: state.routines.map((r) => ({ ...r, isActive: r.id === id, updatedAt: new Date().toISOString() })),
```

`setActiveRoutine` maps over **every** routine and stamps `updatedAt: now` on each one, whether or not
that routine's `isActive` actually changed.

Any routine whose `capacityChangedAt` map is not yet seeded for every activity type it schedules — a
routine created before the fix and not yet block-edited, or one created through any path other than
the four block mutations — falls back to `routine.updatedAt` in `getCapacityChangedAt`. That value has
just been reset to now, so its forecast evidence collapses: issue #7's failure mode with no block
touched.

`docs/REVIVAL_PLAN.md:117` says *"Once the map exists, routine-level edits no longer touch
confidence"* and names only `updateRoutine` as the remainder. This is a second instance of the same
class, and the plan should say so.

Reachability today is narrow: the routine switcher in `src/components/debug/DebugPanel.tsx:304` is
`__DEV__`-gated, but `src/screens/RoutineScreen.tsx:166-167` reaches it in production — delete the
active routine, then create one — which stamps every *other* routine's `updatedAt` and primes the
problem for any routine reactivated later. The `Routine` type's own doc comment mentions Vacation
Mode, so multi-routine switching is intended near-term functionality.

## Deliver

1. Fix `setActiveRoutine` to touch only the routines whose `isActive` actually changes, and to stamp
   the map rather than only `updatedAt` where a real capacity change is implied. Activating a
   different routine genuinely does change which capacity applies — decide what that means for
   evidence and write the decision down.
2. Do the same audit for `updateRoutine` (the remainder the plan already names): a rename or an
   activate must not touch confidence.
3. **Enumerate every writer of `routine.updatedAt` and of `routine.blocks` in the store and state
   the verdict for each in your report.** Two were missed by inspection already; a list is the only
   way to know the third does not exist.
4. Tests for `setActiveRoutine` and `updateRoutine`, which the capacity-timestamp suite currently
   never calls at all.
5. Add coverage for `duplicateRoutine` carrying `capacityChangedAt`. It is correct today — the spread
[...]

Issue #18, opened 2026-09-11
Quarantine side-car can silently rotate out a genuine generation


RogueKostas:
## Why

Found in the non-author review of PR #17 (F1). **P1, CONFIRMED.**

`merge()` in `src/store/useAppStore.ts` only calls in-memory `set()`; it never persists. The
quarantine path is therefore merge-only, and the branch's own test in
`tests/store/hydrationAndBackup.test.ts` pins this by asserting
`AsyncStorage.getItem(APP_STORAGE_KEY)` is byte-identical to the raw blob after a successful
quarantine-and-hydrate.

So the corrupt record stays on disk. Every cold start before some *other* store-mutating action
re-quarantines the identical record into a new generation via `appendQuarantineGeneration`
(`src/store/persistence.ts:126-137`). With `MAX_QUARANTINE_GENERATIONS = 20` and
`.slice(-MAX_QUARANTINE_GENERATIONS)`, the **oldest** generation is silently dropped once the cap is
exceeded.

Failure scenario: a user hits one quarantine event for record A. Later, record B goes bad and is
never cleaned from the blob. Twenty cold starts later — routine on mobile, where the OS kills
backgrounded apps, and likelier still for someone who mostly reads their stats without starting a
timer — generation 1 containing record A is gone, silently.

The comment at `persistence.ts:64-66` calls the cap "realistically unreachable". The test asserting
the blob is left unchanged is precisely what makes it reachable. That contradiction is the finding.

## Deliver

Pick one and say why in the code:

1. Persist the cleaned blob after a merge-only quarantine, so a record is set aside exactly once —
   but only after the side-car write has succeeded, never before (see #19), and without
   reintroducing the first-launch overwrite risk R1 was about.
2. Make the side-car idempotent: key generations by record identity or content hash so re-seeing the
   same bad record does not consume a generation.
3. Keep the cap but never let a full side-car silently discard; surface it.

Whichever is chosen, correct or remove the "realistically unreachable" comment — it is now known to
be wrong.

## Gates

`npm ci` from the tracked lock first (a fresh worktree has none), then `npm run typecheck`,
`npm test`, `npm run build:web`, each in its own process. A negative control per new test: RED
before, GREEN after, both pasted. Suite was **72** at the merge of PR #17; it must not go down.

## Do not

Do not push, open a PR, merge, or run `gh auth`, `git stash`, `git reset --hard`, `npm audit fix`,
[...]

Issue #11, opened 2026-09-11
Repoint the Render blueprint at main, and account for the monthly spend


RogueKostas:
## Why
`render.yaml` defines the blueprint `zenroutine-beta` with the static site `zenroutine-web`, built from `RogueKostas/ZenRoutine`. The revival plan records it as sourced from the `codex/r1-data-safety` branch, not `main`. PR #2 has now merged that branch, so the beta is building from a branch that is no longer where work happens.

Separately, the director sees roughly USD 17/month leaving his account for Render and is not certain what it buys. A Render static site is normally free; USD 17 suggests other services on the account, or a paid instance type.

## Deliver — measure, then recommend. Do not change anything on Render.
1. What `zenroutine-web` is actually configured as, and what it costs.
2. What else exists on that Render account. Other projects of the director's are explicitly out of scope for any change.
3. Whether repointing the blueprint to `main` is safe now, and what the auto-deploy trigger (`checksPass`) will do once it is.

## Do not
Do not modify, delete, or create any Render resource. Do not touch anything belonging to another project. Report only.

## Done means
A short written answer to "what am I paying for, and should the beta now build from main", with a recommendation the director can approve in one word.

RogueKostas:
**Director decision, 2026-09-11:** *"in general want to build from main"* — and the Render
subscription itself gets looked at tomorrow.

So this issue splits in two.

### Repo side — mine, queued as a lane

`render.yaml` declares no `branch`, which is why the blueprint is still following
`codex/r1-data-safety` (the branch it was created from, now merged and deleted from the workflow).
The declarative fix is to state `branch: main` on the `zenroutine-web` service so the intent lives
in the repository rather than only in the Render dashboard.

Queued behind the three lanes currently in flight (#6, #23, #10) — it is a one-line config change
and will go out with a gate like anything else. **Nothing on Render will be touched by that lane**;
the deny list forbids it and the brief will say so.

### Render side — yours, tomorrow

Two things, and the first is the one that actually matters:

1. **Re-sync the blueprint** so `zenroutine-web` follows `main`. A `render.yaml` change alone does
   not move a service that was created against another branch; Render has to pick the change up.
2. **The spend.** A static site is normally free, so ~USD 17/month suggests either a paid instance
   type on this service or other services on the account. Worth knowing which before it renews.

### One consequence worth seeing before you flip it

`render.yaml` sets `autoDeployTrigger: checksPass`. Pointed at `main`, that means **every merge
[...]

PR #27, opened 2026-09-12
docs: replace inherited test counts with measured ones, close D1-D8 (#10)

### LINKED ISSUE #10: Revival plan is stale: says 11 tests, the suite is 62; fold in the product decisions
## Why
`docs/REVIVAL_PLAN.md` records "all 11 core/store tests" in its SDK 55, 56 and 57 evidence sections. Measured at merge time on the branch tip: **7 files, 62 tests**. An inherited number repeated as fact is exactly the failure mode the working agreement warns about.

Separately, `docs/PRODUCT.md` now records the director's answers to the five product questions the plan was waiting on (D1–D8). The plan still presents them as open.

## Deliver
1. Correct every stale test count, naming the tip the new count was measured at. Do not delete the old claim — state what replaced it.
2. Mark the five product questions answered and link `docs/PRODUCT.md`.
3. Reflect the decisions in the milestones: R2 gains the two tracking modes and the weekly-review ritual, R3 gains the tappable forecast breakdown, R5 stops being conditional because accounts and sync are now planned work.
4. Add the F1–F7 findings to the plan's assessment section so the next reader does not rediscover them.

## Do not
Do not touch source or tests. Documentation only. Do not push or open a PR.

## Done means
No number in the plan is unbacked by a run it names, and a cold reader can see which product questions are closed.

RogueKostas:
Closes #10.

Lane `docs-truth`, docs only — no source or test file was touched, and typecheck and `build:web`
were re-run after the edits to prove it.

## Every number replaced, with the run behind it

| Was | Where | Now |
|---|---|---|
| 11 tests | R0 evidence | 79 |
| "all 11 core/store tests" ×3 | SDK 55 / 56 / 57 | 79 |
| "54 tests across six files" | R1 | 7 files, 79 (and **62** at its own tree) |
| "54/54" timezone run | R1 | 62 at that tree |
| 16/16 focused R3 gate | R3 | 19 |

All from `npm test` / `vitest --reporter=verbose` at `9cca4f2`, Node v22.23.2, Vitest 4.1.11.

## Two things the issue did not know

**The "11" was never wrong.** At `f228fd1` and `a192181` the suite genuinely is `engine.test.ts` (8)
plus `useAppStore.test.ts` (3), and `git log eb357c3..a192181 -- tests` is empty — no test changed
anywhere across SDK 54→57. The defect is **tense, not arithmetic**: three sections re-recorded a true
figure in a form that reads as a description of the suite. They are dated rather than "corrected",
which is the right treatment for a record.

**R1's "54 across six files" was wrong at the commit that claims it.** `d5b1e87` has *seven* test
files, and 54 is exactly the static count of `it(` declarations. The runtime count there is 62 —
`persistence.test.ts` has two `it.each` blocks expanding to eight more cases than are written. The
same +8 gap holds today: 71 declarations, 79 cases. Someone counted declarations and wrote it down as
a run. That is a better finding than the one I briefed.

The lane also re-verified the #8 closure against the installed tree rather than the changelog alone:
`babel-preset-expo@57.0.10`, `build/configs/expo.js:92` reads `options.transformImportMeta !== false`
— opt-out, so default-on. Stronger evidence than the changelog line.

## New: an R6 section

D7's AI-subscription workstream had no milestone anywhere and was going to fall between
`docs/PRODUCT.md` and the roadmap. It is now recorded, explicitly marked **unscoped, not planned
work**.

## What the lane could not do — my fault, and worth fixing

**`gh` is not in the lane allow list**, so it could not read the issue tracker and could not confirm
the follow-ups #18–#24. Rather than invent them it recorded the substantive remainders from committed
sources — the two fix commits' own NOT OBSERVED blocks — and flagged in the document that the mapping
to those issue numbers is unverified. F1–F6 → #3–#8 *is* attested by the merge commits and the state
file; **F7 → #9 is the lane's inference and is labelled as such.** I am adding read-only `gh issue
view` to the lane allow list so a documentation lane can cite the tracker.

## Outside its scope, raised anyway — correctly

- **The public beta is serving code missing both merged fixes.** The Render blueprint still builds
  from `codex/r1-data-safety` while R1–R3 plus #3 and #7 are on `main`. Pointing a reviewer at that
  URL right now earns bug reports for things already fixed. That is #11, where the director has
  since decided to build from `main`; the Render-side re-sync is his.
- **#4, #5, #6, #9 are still open on `main`** and the first three are data-loss or wrong-output
  defects. (#6 and #9 land in the PR alongside this one.)

## NOT OBSERVED

R1's forced-timezone `America/Los_Angeles` claim was **not** re-run and is now marked unverified in
the document rather than carried as fact.


RogueKostas:
Orchestrator review, non-author. Gate run by me on the merged tree at `15c0a76` (0 commits behind origin/main): `npm ci` 0, `typecheck` 0, **86 tests**, `build:web` 0. `git diff --name-only origin/main...HEAD` matching `^src/|^tests/` returns **0** — documentation only, as claimed. Gated push: 0 deletions before and after sync, remote ref read back and matching.

The distinction this lane drew is the part worth keeping: **'accurate when recorded, now stale' is not the same defect as 'wrong when recorded'**, and the document now separates them instead of flattening both into a correction. The 11 was true at the commit that wrote it; the 54 never was, because it counts `it(` declarations and two `it.each` blocks expand to eight more cases than are written. Finding that took reading the tree, not the diff.

One thing this PR cannot fix, and it is the same disease in miniature: the counts inside it were measured at `9cca4f2`, and main is now at **86** after #6 and #23 landed while this lane was running. That is not an error — every figure is cited to the commit it was taken at, which is exactly the contract asked for, and a record that dates its measurements does not go wrong when the world moves. It does mean the plan trails main by design. Recording the current figure here so the chain is unbroken: **86 tests at `e79ed46`**, measured by me.

`gh` not being in the lane allow list was my omission — the lane could not read the tracker and correctly refused to invent #18–#24, labelling the mapping unverified rather than guessing. Read-only `gh issue view` is going into the allow list so the next documentation lane can cite the tracker.

Merging.

PR #28, opened 2026-09-12
fix(store): gate the quarantine pointer-clear on the blame relationship (#20)

### LINKED ISSUE #20: Pointer-clear gate does not check the blame relationship, masking unrelated corruption
## Why

Found in the non-author review of PR #17 (F1). **P2, CONFIRMED.**

`src/store/persistence.ts:536-554` clears a dangling `currentTrackingEntryId` when
`currentTrackingEntryId !== null && quarantine.length > 0 && !entryIds.has(currentTrackingEntryId)`.

It never checks that the dangling id has anything to do with what was quarantined — only that
*something* was quarantined on this run.

Failure scenario: entries are `[E1]`, `E1` is closed, nothing is open. A different, unrelated entry
is malformed and gets quarantined. `currentTrackingEntryId` holds a stale value that never referred
to any real entry — from an unrelated bug, a bad manual edit, a partial write. Because something
else was quarantined, the pointer is silently nulled, the strict invariant at the end passes, and
hydration reports success while genuinely unrelated corruption is swallowed.

Before PR #17 this threw `Invalid currentTrackingEntryId` and surfaced the corruption.

The comment two lines above the condition says *"a dangling pointer with no bad record to blame is
real corruption and should still be loud."* That is the intent; the code does not implement it.
`tests/store/persistence.test.ts` covers only `quarantine.length === 0`, so the case is untested.

## Deliver

1. Gate the clear on the blame relationship — the dangling id must match a record that was actually
   quarantined — so an unrelated dangling pointer stays loud, as the comment already promises.
2. Tests for both: quarantined-entry-was-the-running-timer clears the pointer; quarantine happened
   but the pointer dangles for an unrelated reason still throws.

## Gates

`npm ci` from the tracked lock first, then `npm run typecheck`, `npm test`, `npm run build:web`,
each in its own process. A negative control per new test: RED before, GREEN after, both pasted.
Suite was **72** at the merge of PR #17; it must not go down.

## Do not

## Done means

The code does what its own comment says, and both branches are pinned by a test.

RogueKostas:
Closes #20.

Lane `quarantine-blame`, headless `claude -p` in its own worktree from `1fbc197`.

## What it does

PR #17 cleared a dangling `currentTrackingEntryId` whenever `quarantine.length > 0` — without ever
checking that the dangling id had anything to do with what was quarantined. A stale pointer that
never referred to any real entry was therefore silently nulled whenever *something else* went into
quarantine, swallowing unrelated corruption that used to throw. The comment two lines above said a
dangling pointer with no bad record to blame should stay loud; the code did not implement it.

Blame is now two-armed (`src/store/persistence.ts:562-592`):

- `dropped.id === pointer` — the entry the timer named is exactly what was dropped, or
- `dropped.id === null` — the record's own id is what failed to parse, so it cannot be ruled out.

A dropped record with a *readable* id belonging to some other entry now explains nothing, and the
pointer stays loud.

## The lane's correction to my brief — and it changed the fix

**My brief said the `quarantine.length > 0` case was untested. It is not.**
`tests/store/persistence.test.ts:194`, pre-existing from PR #17, covers a non-empty quarantine: the
open entry's own `id` is unreadable, it is dropped as `id: null`, and the pointer is cleared. What
was genuinely untested is the *unrelated* dangling pointer with a non-empty quarantine — the
scenario the issue is actually about.

That distinction is what produced the two-armed rule. The naive reading of my own acceptance
criterion — "the dangling id must match a record that was actually quarantined", i.e. an id-set
membership test — **would have broken that existing test**, because a record whose `id` is the
corrupted field can never match anything. The hazard PR #17's comment named was real; only its
conclusion was wrong.

My brief's line numbers (`536-554`) were also stale. The block was at `564-581` before the change.

| Gate | Lane |
|---|---|
| `npm ci` from tracked lock | exit 0, 584 packages, lock confirmed tracked and not regenerated |
| `npm run typecheck` | exit 0, no new `any` / `@ts-ignore` |
| `npm test` | exit 0, **88 passed** (86 at `1fbc197`, +2) |

The lane re-measured the 86 at its own tree before touching anything rather than trusting the brief.

Two negative controls, one per new test, both RED then GREEN. The second is the one that matters:
reinstating the old `quarantine.length > 0` gate produced `AssertionError: expected [Function] to
throw an error` — **that RED is the bug reproduced.**

## NOT OBSERVED

- **#15 is open: no physical-device smoke has ever been run on this project.** This touches the
  hydration path, which is precisely where device-only behaviour lives.
- Unit-level coverage of `migratePersistedState` alone. Nothing exercised through `useAppStore.ts`
  or against real AsyncStorage — the lane stayed out of that file because another lane is in it,
  and no change there was required.
- **A pre-existing widening the lane found and deliberately did not touch:** `pendingQuarantine` is
  shared across the `migrate` and `merge` stages of a single hydration run and is reset only in
  `onRehydrateStorage`. So on a legacy blob, blame evaluated during `merge` can be satisfied by a
  record dropped during `migrate`. Still far narrower than the old gate, but not per-stage. Folding
  it into #5, which is already in this code.
- The `id: null` arm is deliberately permissive — any unreadable-id record vouches for any dangling
  pointer on that run. That is the honest bound available without inventing identity for a record
  whose identity is the corrupted part, and the lane flagged it as a judgement someone should agree
  with. I agree with it: the alternative is to brick a user over a record we cannot name.


RogueKostas:
Orchestrator review, non-author. Gate run by me on the merged tree at `beb716d` (0 commits behind origin/main): `npm ci` 0, `typecheck` 0, **88 tests**, `build:web` 0. Gated push: 0 deletions before and after sync, remote ref read back and matching.

I read the diff. The predicate is right and, more to the point, the *reasoning* is now in the file: the comment explains why this is not `quarantine.some(d => d.id === pointer)`, which is the trap my own acceptance criterion would have walked into.

**The lane's correction is the substance of this PR.** I wrote that the non-empty-quarantine case was untested. It is tested — `tests/store/persistence.test.ts:194`, from PR #17. What was untested is the *unrelated* dangling pointer, which is the actual issue. Had the lane implemented my criterion literally, it would have broken that existing test, because a record whose `id` is the corrupted field is quarantined as `id: null` and can never match an id set. PR #17's comment named that hazard correctly and only drew the wrong conclusion from it. My brief's line numbers were stale too.

I agree with the judgement on the `id: null` arm being permissive. Any unreadable-id record vouches for any dangling pointer on that run, which is looser than one would like — but the alternative is bricking a user over a record we cannot name, and that is the failure this whole line of work exists to remove.

The pre-existing widening the lane found and left alone — `pendingQuarantine` shared across the `migrate` and `merge` stages of one hydration run, reset only in `onRehydrateStorage` — is real and is going into #5, which is already in this code. Correct call not to widen scope for it.

Merging.

PR #30, opened 2026-09-12
fix(store): stop routine-level writes collapsing forecast confidence (#24)

### LINKED ISSUE #24: setActiveRoutine and updateRoutine bump updatedAt without stamping the capacity map

RogueKostas:
Closes #24.

Lane `routine-writers`, headless `claude -p` in its own worktree. It did the thing the issue actually
asked for — an exhaustive audit — and the audit found more than the issue did.

## The rule the three fixes share

`getCapacityChangedAt` falls back to `routine.updatedAt`. Therefore: **any writer that bumps
`routine.updatedAt` must also seed `capacityChangedAt` via `withCapacityChangedAt`**, or the bump
resets every un-seeded activity type. That generalisation is what the plan was missing, and it is now
written into `docs/REVIVAL_PLAN.md`.

## Every writer of `routine.updatedAt` / `routine.blocks`, with a verdict

Fifteen writers enumerated. Three were broken and are fixed here — `updateRoutine`,
`setActiveRoutine`, `duplicateRoutine`. One more is a real defect of the opposite polarity and is
**deliberately left**, see below. The rest are sound, most because they create routines with zero
blocks so nothing can fall through the fallback.

## Two corrections to my brief

**1. I said `duplicateRoutine` "is correct today" and only needed test coverage. Wrong.** The spread
does carry `capacityChangedAt` and nothing mutates it in place — both true, both beside the point.
The copy also sets `updatedAt: now`, so any activity type the *source* never stamped falls through
onto that fresh timestamp and **the copy is born with zero evidence for a schedule identical to the
source's.** That is a fourth instance of the bug class, not a coverage gap. Fixed, not just covered.

**2. There is a third writer I did not name: `deleteGoal`** (`useAppStore.ts:535`) rewrites
`routine.blocks` to clear `goalId` — a `CAPACITY_RELEVANT_BLOCK_FIELDS` field — and stamps nothing.
It is the opposite polarity to #7: it cannot collapse confidence, it silently **under-reports** a
real capacity change.

The lane left it unfixed on purpose, and I agree with the reasoning: whether deleting a goal should
reset its *competitors'* evidence is a product decision, not a mechanical fix, and fixing it inside
this lane would have been the lane making that call silently. Recorded in `REVIVAL_PLAN.md` and
filed as its own issue.

## The activation decision, which deserves your eye

Activating a routine *is* a capacity change — but only for the activity types whose schedule
genuinely differs from the routine going out, compared per activity type by a signature derived from
`CAPACITY_RELEVANT_BLOCK_FIELDS` (so a field opted into that list is compared here too). Switching
between routines that schedule Fitness identically leaves Fitness's evidence intact; a routine that
halves it resets it. The routine being *deactivated* stamps nothing — its blocks did not move, and
nothing reads it until it is activated again, which compares afresh. `outgoing` resolves via
`activeRoutineId` rather than `isActive`, because that is what `useActiveRoutine` reads.

**The consequence worth arguing about:** a Vacation Mode round trip *will* reset the confidence of
the activity types Vacation schedules differently, because evidence gathered on holiday was not
gathered under the work schedule. Someone may reasonably want the opposite. The comparison is
isolated in `activationCapacityChanges`, so reversing it is a one-function change.

## Beyond the brief, and right

`updateRoutine` now takes `RoutineUpdate`, which excludes `capacityChangedAt` **and `isActive`**. The
second exclusion is the lane's own: `isActive` has a companion, `activeRoutineId`, which is what
`useActiveRoutine` actually resolves and which only `setActiveRoutine` maintains, so
`updateRoutine(id, {isActive: true})` would have left the two disagreeing about which routine a
forecast reads. No call site passed either field. The runtime matches the type via an explicit
`UPDATABLE_ROUTINE_FIELDS` allow list, so a plain-JS caller cannot smuggle a payload through
`Object.keys` either.

| Gate | Lane |
|---|---|
| `npm ci` from tracked lock | exit 0, confirmed tracked, not regenerated |
| `npm test` | exit 0, **96 passed** (86 at its own re-measured baseline, +10) |
| `git diff --name-status origin/main...HEAD` | 3 `M`, no `D` |

Two `@ts-expect-error` directives are added **deliberately**, as the type-level check that
`capacityChangedAt` and `isActive` are rejected — they fail the typecheck if the type ever loosens.

Ten new tests, three deliberate breaks, each reverted: restoring pre-fix `setActiveRoutine` reddens
5, pre-fix `updateRoutine` reddens 3, and `capacityChangedAt: undefined` in `duplicateRoutine`
reddens 2. The lane also noted honestly that dropping only the *seeding* call would not have
reddened the "carries the map" test, because the spread still carries it — that test guards the
refactor case, and the seeding is guarded by the other one.

## NOT OBSERVED

- **#15 is open — no physical-device smoke has ever been run on this project.**
- **Two of the three fixes are unreachable by a user today.** `updateRoutine` and `duplicateRoutine`
  have **zero call sites** in the app; their fixes are store-API-only. Only `setActiveRoutine` is
  reachable — `DebugPanel.tsx:304` (`__DEV__`) and `RoutineScreen.tsx:166` — and the lane confirmed
  the production path is benign under the fix, because the routine it activates is brand-new and
  empty.
- The 7/14 distinct-day confidence thresholds remain the unvalidated prototype assumption. Every test
  here asserts the *cutoff timestamp* is right; none can say whether the label a user finally sees is.


RogueKostas:
Orchestrator review, non-author. Gate run by me on the merged tree at `41c01b0` (0 commits behind origin/main): `npm ci` 0, `typecheck` 0, **98 tests**, `build:web` 0. Gated push: 0 deletions before and after sync, remote ref read back and matching.

This is the lane doing more than it was asked and being right about all of it.

**Two corrections to my brief, both accepted.** I wrote that `duplicateRoutine` was correct and only needed coverage — it was not: the copy sets `updatedAt: now`, so any activity type the source never stamped falls onto that fresh timestamp and the copy is born with zero evidence for an identical schedule. A fourth instance of the bug class, not a coverage gap. And the third writer I could not name exists: `deleteGoal` rewrites `routine.blocks` clearing `goalId` and stamps nothing.

**Leaving `deleteGoal` unfixed was the right call** and I want that on the record, because the easy thing would have been to patch it and move on. Whether deleting a goal should reset its competitors' evidence is a product decision — filed as #31 for the director, with options and a recommendation, rather than decided silently inside a lane.

**The signature tightening beyond the brief is the best judgement in this PR.** Excluding `isActive` from `RoutineUpdate` is not what I asked for, and it is correct: `isActive` has a companion in `activeRoutineId`, which is what `useActiveRoutine` actually resolves and which only `setActiveRoutine` maintains, so `updateRoutine(id, {isActive: true})` would have left the two disagreeing about which routine a forecast reads. The `UPDATABLE_ROUTINE_FIELDS` allow list closes the plain-JS route too, and the two `@ts-expect-error` directives make the type itself the test.

**The activation semantics are a product judgement and the PR says so.** Stamp only the activity types whose schedule genuinely differs from the routine going out. A Vacation Mode round trip will therefore reset confidence for the types Vacation schedules differently. I think that is right — evidence gathered on holiday was not gathered under the work schedule — but it is arguable, and the comparison is isolated in `activationCapacityChanges` so reversing it is one function.

Noting the honest scope limit the lane volunteered: two of the three fixes are store-API-only, because `updateRoutine` and `duplicateRoutine` have zero call sites in the app today. Only `setActiveRoutine` is reachable, and its production path is benign under the fix.

Merging.

Issue #19, opened 2026-09-11
MEASURE: can zustand's migrate write-back lose a quarantined record on a pre-v4 blob?


RogueKostas:
## Why

Found in the non-author review of PR #17 (F1). **P1, PLAUSIBLE — not traced. This issue is to
MEASURE, not to fix.**

`src/store/useAppStore.ts` wires the quarantine sink into both `migrate` and `merge`. Zustand's
`persist` middleware calls `migrate()` only when the on-disk `version` differs from
`options.version`, and — per its documented behaviour — writes the migrated state back to storage
*before* `rehydrate()`'s promise resolves. The quarantine side-car append happens later, in
`initializeAppStore`.

If that ordering is real, then on a device still holding a **pre-v4 blob** containing one
structurally malformed tracking entry:

1. `migrate()` quarantines the record and returns clean state.
2. Zustand immediately persists the clean, already-stripped blob.
3. The side-car append runs afterwards — and its failure path is only `console.warn`.

A failed side-car write, or the process being killed in the gap, loses the record from **both**
locations, permanently and silently. That is the one outcome the quarantine design exists to prevent.

## This is a measure-and-report task. Do not fix anything.

Determine, with evidence:

1. What `zustand@5.0.11`'s `persist` actually does. Read the installed
   `node_modules/zustand/esm/middleware.mjs` (or the CJS sibling) after `npm ci` — the source, not
   the documentation. Does a version mismatch trigger a `setItem` before `rehydrate()` resolves?
2. Whether the window is real in this app: write a pre-v4 blob containing a malformed tracking
   entry, hydrate, and observe the order of storage writes. Assert on the observed order rather than
   describing it.
3. If the window is real, how wide it is and what the cheapest correct ordering would be — the
   side-car write succeeding before anything rewrites the app blob is the obvious candidate.

`NOT OBSERVED` is an acceptable outcome. Say plainly what you could not determine.

## Do not

Do not modify source, do not commit, do not push. Report only. Label every piece of evidence with
the commit it came from.

## Done means

A written verdict on whether a pre-v4 device can lose a quarantined record entirely, with the
zustand source and an observed write order behind it, and a recommendation.

RogueKostas:
**Verdict: CONFIRMED. Closing this issue; the fix belongs to #18.**

Lane `zustand-probe`, forbidden from fixing anything, delivered the measurement. Worktree verified
[...]

PR #32, opened 2026-09-12
fix(store): stop the quarantine side-car losing a record entirely (#18)

### LINKED ISSUE #18: Quarantine side-car can silently rotate out a genuine generation

RogueKostas:
Closes #18.

Lane `sidecar-durability`. Both defects in that issue — the ordering window CONFIRMED by #19, and
the generation rotation the issue was originally filed for.

## Ordering

`migrate` is now async and awaits the side-car write before returning clean state. A failure
propagates and aborts the rehydrate chain **before** zustand's `setItem`, so the blob keeps the
record. The lane verified zustand 5.0.11's source itself rather than taking #19's word for it:
`middleware.mjs:392-396` chains a promise returned from `migrate` ahead of the `set` + `setItem` at
`:419-421`, and `:405` short-circuits the non-migrate path with no write at all.

The merge path still swallows failures, because nothing rewrites the blob there. That asymmetry is
the point and is commented as such.

## Rotation — option 2, not option 1, and the reason is good

The issue offered three options. The lane took **an idempotent side-car** over **cleaning the blob**,
and said why in the code: *a fix that only works when the cleanup write lands does not cover its own
worst case, which is precisely a device that cannot write.* Dedup is by content fingerprint over the
whole generation, so a genuinely new bad record still gets its own generation, and the two write
sites — the migrate path and `initializeAppStore` — no longer double-append.

## A behaviour change that is not a pure bug fix

On the migrate path, an unwritable side-car now makes the app **refuse to open** (retryable hydration
error) rather than open having destroyed the record. The lane judged refusing the lesser harm when
the device holds the only copy, and flagged it rather than burying it. I agree — silent data loss is
worse than a retryable failure — but it is a real change in what a user with a full disk sees, and it
has never been exercised against real storage.

| Gate | Lane |
|---|---|
| `npm ci` from tracked lock | exit 0, lockfile unchanged |
| `npm test` | exit 0, **103 passed** (98 at its own re-measured baseline, +5) |

**Control A** (sync `migrate`) reproduced #19's loss verbatim — blob rewritten to
`"trackingEntries":[],"version":4`, side-car write rejected, a `console.warn` the only trace:
`expected 3 to be less than 1` (side-car write at 3, blob write at 1). **Control B** (dedup off)
reproduced #18: `entry-record-a` gone, twenty copies of the same record. Both restored.

Per the trap warning, the tests use an invalid `source`, not `endTime < startTime` — the lane
confirmed the latter is repaired on a v3 envelope and would have passed vacuously.

## The residual, decided and explained

Left in place, deliberately. `merge` cannot take the `migrate` treatment because zustand calls it
**synchronously** (`:416`) — there is no await to hang the write on. So it stays safe by ordering
*within* `initializeAppStore`: the side-car write precedes `persistSnapshot`, the first thing that can
rewrite the blob. That is now explicit and commented, including a warning that a store write added
above that point reopens the window. It is a convention, not a compiler-enforced constraint.

## Owed elsewhere

The issue asked for the `realistically unreachable` comment at `persistence.ts:63-66` to be corrected
or removed. **This lane could not**, because #5's lane held that file. It documented why the comment
is wrong at its own call site instead. Carried into the follow-up issue.

## NOT OBSERVED

- **#15 is open: no physical-device smoke has ever been run**, and this is the hydration path.
  Everything is verified on the vitest Map-backed AsyncStorage mock and the web export. That matters
  more than usual here, because the new refuse-to-open failure mode has never met real storage.
- Write *order* only; no wall-clock measurement. Kill-during-the-gap not simulated — the demonstrated
  loss is the deterministic failed-write path.
- The lane also disclosed a harness bug of its own: `restoreMocks` does not reset a `vi.fn(impl)` base
  implementation, so its first version leaked a rejecting mock into later tests. Found and fixed
  before commit.


RogueKostas:
Orchestrator review, non-author. Gate run by me on the merged tree at `fcde340` (0 commits behind origin/main): `npm ci` 0, `typecheck` 0, **103 tests**, `build:web` 0. Gated push clean, remote ref read back matching. **This closes the last p0 in the backlog.**

Three things I want on the record.

**The lane verified zustand's source itself instead of trusting #19.** The brief handed it a conclusion and it re-derived it. That is the correct instinct for a fix whose whole value is an ordering guarantee.

**It chose option 2 over option 1 and the reasoning is better than my issue's.** I listed cleaning the blob first. The lane took the idempotent side-car because *a fix that only works when the cleanup write lands does not cover its own worst case, which is precisely a device that cannot write.* That is right, and it is the kind of reasoning I would not want a lane to skip in order to follow a brief.

**The behaviour change is real and is flagged, not buried.** On the migrate path an unwritable side-car now makes the app refuse to open — retryable — rather than open having destroyed the record. Silent data loss is worse than a retryable failure, so I agree with the call, but a user with a full disk now sees something new, and that path has never met real storage. It is in the NOT OBSERVED list and in #15's shadow.

The residual is left with an explanation rather than a shrug: `merge` cannot take the same treatment because zustand calls it synchronously, so it stays safe by ordering within `initializeAppStore` — now explicit and commented, including the warning that a store write added above that point reopens it. A convention, not a constraint, and the PR says so.

The `realistically unreachable` comment at `persistence.ts:63-66` is still owed; #5's lane held that file. Carried into the follow-up.

Merging.

PR #33, opened 2026-09-12
fix(app): scope quarantine dismissal to the event, and cover the notice (#21)

### LINKED ISSUE #21: Quarantine notice: dismissal is not scoped to the event, and the side-car is write-only
## Why

Found in the non-author review of PR #17 (F1). **P2, CONFIRMED.**

`App.tsx:56-60` and `App.tsx:143-145`: `quarantineDismissed` is a bare `useState(false)`, only ever
set to `true`, and never reset except by unmounting `AppContent`.

So if a user dismisses one quarantine notice and a later hydration in the same mounted instance —
a forced rehydrate, or `resetAppStoreAfterHydrationError` followed by another problem — quarantines a
**different** record, `showQuarantineNotice` stays `false` for the life of that component. The
side-car write still happens; the user is simply never told.

Related, and worth deciding at the same time: **the side-car is write-only.** Records are retained
under `zenroutine-quarantine` but nothing reads them back, so the notice can say records were set
aside and cannot show them. The lane that built it flagged this as needing a product decision.

## Deliver

1. Scope the dismissal to the quarantine event rather than the component instance, so a later,
   different event is reported.
2. Give the notice automated coverage. It currently has **none** — it is never rendered in any test;
   only typecheck and the web bundle touch it. Layout, the dismiss button and
   `accessibilityLiveRegion` are unverified anywhere. This was named as the weakest part of PR #17.
3. Product decision, to be answered before or alongside this: should quarantined records be
   viewable, and restorable, in the app? A notice the user cannot act on is close to no notice.
   `docs/PRODUCT.md` D6 says local data is the product promise until sync exists, which argues for
   viewable.

## Gates

`npm ci` from the tracked lock first, then `npm run typecheck`, `npm test`, `npm run build:web`,
each in its own process. A negative control per new test: RED before, GREEN after, both pasted.
Suite was **72** at the merge of PR #17; it must not go down.

## Do not

`expo prebuild`, `eas build` or any Render command. Corrections to this issue are welcome and
expected.

## Done means

A second, different quarantine event reaches the user, and the notice is exercised by at least one
test rather than by nothing at all.

RogueKostas:
Closes #21's engineering half. The product question it raises is answered with a costing, not a
guess.

## What changed

`quarantineDismissed` was a bare `useState(false)`, never reset, so a later *different* quarantine
event in the same mounted instance was never announced. It now holds **which report** was dismissed,
and a hydration that sets different records aside is a different report, so it is announced.

The banner is extracted to `src/components/common/QuarantineNotice.tsx` with the copy, the live
region, the button role and the 44pt target unchanged from PR #17.

## Two corrections to my brief

**1. The side-car is not write-only in code — only to the user.** I wrote "nothing reads them back".
`readQuarantineArchive()` exists at `useAppStore.ts:87`, `parseQuarantineArchive` is exported and
tested, and the generation cap and record shape are all in place. What is missing is anything that
*surfaces* them. That makes a viewer far cheaper than I implied.

**2. There is no way to render a React Native component in this suite, and getting one would break a
hard rule.** `vitest.config.mts` is `environment: 'node'` with `include: ['tests/**/*.test.ts']`, and
no renderer is in `node_modules` or the lock. Adding one means regenerating the **tracked**
`package-lock.json` — forbidden. So the notice was covered without a renderer, by asserting the
returned element tree with `react-native` mocked to element names. That verifies the props the notice
owns; **it does not verify native rendering**, and the commit message says so.

## Gates

| Gate | Lane |
|---|---|
| `npm ci` from tracked lock | exit 0, confirmed tracked, untouched |
| `npm test` | exit 0, **94 passed** (88 at its own re-measured baseline, +6) |
| `git diff --name-status origin/main...HEAD` | 2 `M`, 2 `A`, no `D` |

Six negative controls, one per new test, each RED then GREEN. The first reinstates the original bug —
`report.length > 0 && dismissedReport === null` — and reds the second-event test exactly.

One test is a deliberate **tripwire on a store contract**: the fix identifies an event by the array
reference the store hands out (fresh per quarantining hydration, one reference-stable empty array
otherwise, forced by `useSyncExternalStore`). If `getQuarantinedTrackingEntries()` ever returns a
fresh array per read, that test goes red — and it should.

## The product question, costed rather than guessed

**Viewing is cheap: roughly half a day, no new dependencies.** Already present — the archive, the
types, a tested parser, the generation cap, and per-entry `index` / `id` / `reason` / `record`.
Missing: one exported store read (~15 lines), UI in `SettingsScreen`, which already has the
export/import modal pattern (~60 lines there, or ~150 as its own screen), a "See what was set aside"
action beside Dismiss (~10 lines), and tests.

**Restoring is a feature, not an increment.** A plain restore is wrong: the reason a record was
quarantined has not changed, so re-inserting it means the next hydration quarantines it again.
Restore really means *repair* — show the record, let the user fix the field that failed validation,
re-validate, insert, drop it from the archive. That is a write path into user data with its own
migration and test burden.

The lane's recommendation, which I endorse: **ship viewing** — `docs/PRODUCT.md` D6 argues for it, and
a notice the user cannot act on is close to no notice — and treat repair-and-restore as a separate,
later decision. Not built here.

## NOT OBSERVED

- **#15 is open: no physical-device smoke has ever been run.**
- **The banner has no safe-area padding.** It renders inside the plain container `View` above
  `NavigationContainer`; `SafeAreaProvider` wraps everything but the banner applies no insets. On a
  notched iPhone or under an Android status bar it may sit beneath the system bar. Unchanged from
  PR #17 — the lane did not alter the layout — but only a device or a browser settles it.
- **`accessibilityLiveRegion` is Android- and web-only in React Native; iOS has no equivalent prop**
  and would need `AccessibilityInfo.announceForAccessibility`. If that is right, the notice is not
  announced on iOS today. The tests assert the prop is set, not that any platform acts on it. Worth
  confirming against the RN 0.86 docs before the notice is called accessible.
- No screenshot of the extracted component exists on any platform.


RogueKostas:
Orchestrator review, non-author. Gate run by me on the merged tree at `88bf24b` (0 commits behind origin/main): `npm ci` 0, `typecheck` 0, **109 tests**, `build:web` 0. Gated push clean, remote ref read back matching.

Two corrections to my brief, both accepted, and the second is the more useful.

I wrote that the side-car is write-only. In code it is not — `readQuarantineArchive()`, a tested `parseQuarantineArchive`, the generation cap and the record shape all exist. What is missing is anything that surfaces them, which makes a viewer far cheaper than I implied and changes the shape of the product question.

And there is genuinely no way to render a React Native component in this suite without adding a renderer, which means regenerating the tracked lockfile — forbidden. Rather than skip the coverage or quietly break the rule, the lane covered what it could without one and **said plainly what that does not prove**: the props the notice owns are verified, native rendering is not. That is the right trade and the right disclosure.

The array-reference tripwire is a nice touch: if `getQuarantinedTrackingEntries()` ever returns a fresh array per read, the test goes red, and it should.

The costing is what I asked for instead of a guess, and it is specific enough to decide on: viewing is about half a day with no new dependencies; restoring is a different thing entirely, because the reason a record was quarantined has not changed — re-inserting it just gets it quarantined again on the next hydration, so "restore" really means "repair", which is a write path into user data with its own migration burden. I endorse the recommendation: ship viewing, decide repair separately. Neither is built here.

Carrying forward unfixed and unhidden: the banner still has no safe-area insets, and `accessibilityLiveRegion` is Android/web-only in React Native, so on iOS the notice may not be announced at all. Both need a device, which is #15.

Merging.

Issue #34, opened 2026-09-12
Decision needed: what happens when a skeleton record (activity type, goal, routine) is unreadable?


RogueKostas:
## Why

Named, tested and deliberately left open by the `linkage-quarantine` lane while closing #5. It is not
a regression — these routes predate all of the quarantine work — but they are **live, permanent
brick routes** and they should not stay unrecorded.

Two of them survive after #5:

1. A **goal** or a **routine block** whose `activityTypeId` no longer resolves throws
   unconditionally, at every version.
2. A **routine block** with a dangling `goalId` throws at current version.

Either one fails hydration permanently, and the only recovery the UI offers wipes all local data —
the same failure #3 was filed for, reached through a different corruption shape.

A test now pins the current strict behaviour, so the gap is documented rather than latent. That test
should go red when this is fixed; treat it as the marker, not as a contract to preserve.

## Why the lane stopped rather than fixing it

The quarantine side-car holds `QuarantinedTrackingEntry` only. Setting a **goal** or a **routine**
aside means either inventing a second side-car with its own shape and generation handling, or
destroying skeleton data the rest of the store is built on. Every tracking entry, every routine block
and every forecast hangs off activity types and goals; a store that hydrates with a goal missing is a
different proposition from one that hydrates with a tracking entry missing.

That is a design decision, and the lane was right not to take it inside a bug fix.

## The question to answer first

**What should the app do when a skeleton record — an activity type, a goal, a routine — is
unreadable?** Three shapes, none obviously right:

- **A second side-car per record type.** Consistent with what exists, most work, and the recovery
  story stays "we set it aside and can show you".
- **Reconstruct a placeholder.** Hydrate with a synthetic "Unknown activity" so dependent records
  survive, and mark it for the user to rename or merge. Cheapest for the user, but it invents data,
  and `activityTypeId` being required is exactly why #5 could not simply clear the reference.
- **Cascade.** Quarantine the skeleton record and everything that depends on it. Honest, and
  potentially removes a large amount of a user's history at once.

## Also owed here

PR #32 could not correct the **`realistically unreachable` comment at
`src/store/persistence.ts:63-66`**, because #5's lane held that file at the time. It is wrong — #18
established the cap is reachable — and it is a one-line fix. Fold it into whatever lane takes this.

## Do not

Do not implement until the shape above is decided and recorded on this issue. `needs-decision`, not
`agent-ready`.

RogueKostas:
[...]

PR #36, opened 2026-09-12
fix(store): route stale tracking-entry references through the quarantine sink (#5)

### LINKED ISSUE #5: F3: dangling activityTypeId throws unconditionally, locking the user out

RogueKostas:
Closes #5.

Lane `linkage-quarantine`. Stale tracking-entry references now go through the quarantine sink instead
of bricking hydration.

## Three corrections to my brief, all accepted

**1. Line numbers stale, and the referent matters.** `persistence.ts:342-352` is now `:489-499` — an
11-line span covering *all three* activity-type linkage checks, not just the entry one.

**2. "Repair for legacy versions by clearing the reference" is structurally impossible here.**
`activityTypeId` is **required** on `TrackingEntry`, `Goal` and `RoutineBlock` — unlike `goalId?` and
`routineBlockId?`. There is nothing to clear it to. So the version-consistency I asked for cannot be
delivered in the literal sense. What the lane delivered instead: **quarantining *is* the repair for
`activityTypeId`, at every version**, because a legacy blob has no better answer available than a
current one. The outcome the issue actually wanted — a legacy store with a dangling activity type
hydrates — is met, with the entry set aside recoverably rather than a reference nulled.

**3. Two brick routes deliberately left, and named rather than hidden.** A **goal** or **routine
block** whose `activityTypeId` has gone missing still throws unconditionally, and a routine block with
a dangling `goalId` still throws at current version. Out of this issue's named scope, and fixing them
needs a design decision no lane should take alone: the side-car holds `QuarantinedTrackingEntry` only,
so setting aside a goal or a routine means either a second side-car or destroying skeleton data. A
test now **pins the current strict behaviour** so the gap is documented rather than latent. Filed as
a follow-up.

## The per-stage question — solved without touching the other lane's file

`pendingQuarantine` is shared across `migrate` and `merge`. The lane made the blame gate per-stage by
snapshotting `quarantine.length` on entry and reading `.slice(quarantineStartIndex)` — scoping the
*read* rather than resetting the array, which keeps the accumulated report the user is shown intact.
Resetting per stage would have needed `useAppStore.ts`, which another lane held, **and would have been
the worse fix.**

## A live second-order bug the widened scope created, caught by its own control

`entryIds` was computed **before** the linkage phase. Since quarantining now removes entries, a
pointer at a quarantined entry would be judged against the pre-linkage set, trip the strict
`currentTrackingEntryId` check, and re-brick the launch for exactly the reason being fixed. Moved to
after the linkage phase. Control C reds on precisely this — it was live, not hypothetical.

## Gates

| Gate | Lane |
|---|---|
| `npm ci` from tracked lock | exit 0, confirmed tracked, never regenerated |
| `npm run typecheck` | exit 0, no new `any` / `@ts-ignore` / `@ts-nocheck` (grep-proven, including no control residue) |
| `npm test` | exit 0, **96 passed** (88 at its own re-measured baseline, +8) |
| `git diff --name-status origin/main...HEAD` | 3 `M`, no `D` |

Four negative-control cycles, production code reverted each time, all RED then GREEN — every new
test covered by at least one. The #19 trap warning was heeded: dangling references are unaffected by
`repairLegacyValues`, and control A proves the legacy assertions are load-bearing.

## Findings raised, not silently absorbed

- **`goals[].loggedMinutes` is not reconciled when an entry is quarantined.** An entry set aside for a
  stale `routineBlockId` can still carry a valid `goalId`, so that goal keeps minutes no surviving
  entry accounts for. Pre-existing from PR #17; this widens the set of records it can happen to. Not
  fixed — whether a user's goal progress should shrink on launch is a product decision. Filed.
- Error message narrowed to the singular, since it is now per-record and appears as a side-car
  `reason`. Repo grepped: no consumer of the old string.
- The strict-without-sink contract is preserved for all three checks, so **backup import still fails
  loudly**.

## NOT OBSERVED

- **#15 is open: no physical-device smoke has ever been run.** All of this is vitest against a mocked
  AsyncStorage. The torn-write, hand-edit and restore paths that actually *produce* these blobs are
  entirely unobserved.
- **No real corrupted blob was tested.** Every fixture is synthetic. Whether field corruption takes
  these three shapes, or a fourth nobody has enumerated, is unknown.
- The reasons this emits are user-facing strings that reach the notice from #21. Nobody has checked
  they read sensibly to a non-technical user.
- `quarantineStartIndex` assumes the sink is append-only across stages. If `pendingQuarantine`'s
  lifecycle changes, that assumption needs re-checking.

---

## Second dispatch: the conflict, and what its controls proved

The gated push refused this branch the first time — a real content conflict in
`tests/store/hydrationAndBackup.test.ts` with PR #32 (`sidecar-durability`), which had landed in the
same file minutes earlier. **I re-dispatched the lane to resolve its own conflict rather than
resolving it myself**; the author owns the resolution, the orchestrator lands it.

**`persistence.ts` had no conflict at all.** PR #32 was confined to `useAppStore.ts`, so this lane's
work carried across byte-identical. The conflict was purely that both sides appended a new top-level
`describe` at the same offset, and git folded two shared context lines into the hunk even though they
belong to a different `it` on each side. Both blocks are kept in full, neither edited.

**The append-only assumption survived**, which was the real question I sent it back with: PR #32 made
`migrate` async and the side-car idempotent, and `quarantineStartIndex` depends on the sink being
append-only across stages. It is, because #32 never touched the sink's lifecycle.

The control that matters: disabling the `goalId` quarantine reds
`hydration of a store whose references have gone stale > opens, sets the stranded entry aside, and
gets it to the side-car` — **the conflict-resolved test itself**. So the resolution is load-bearing,
not a test reassembled into something vacuous. That was the specific risk of hand-resolving a
test-file conflict and it is the thing the lane most wanted proven. Six controls in total, each a
production-code edit reverted by its exact inverse (`git checkout -- <path>` is denied in a lane
worktree), restore verified by an empty `git diff`.

Suite at the resolved tree: **111**.

## The thing a human should actually look at

The lane raised it and I am repeating it rather than burying it in a list: **PR #32 and this PR are
two independent fixes to the same durability path, merged without either author seeing the other's
code.** The suite and the build agree they compose. Whether the side-car write-ordering and the
linkage quarantining interact correctly on a *real* torn write — the event both exist to survive — is
unobserved, and cannot be observed without a device. It is the sharpest single item behind #15.


RogueKostas:
Orchestrator review, non-author. Gate run by me on the merged tree at `fb0130a` (0 commits behind origin/main): `npm ci` 0, `typecheck` 0, **117 tests**, `build:web` 0. Gated push clean, remote ref read back matching.

This branch was refused by the gated push on its first attempt — a real conflict with PR #32 in `tests/store/hydrationAndBackup.test.ts`. I sent it back to its own author rather than resolving it myself, with one question attached: does #32's async `migrate` and idempotent side-car break the `quarantineStartIndex` append-only assumption? It does not — #32 was confined to `useAppStore.ts`, so `persistence.ts` carried across byte-identical and the conflict was purely two `describe` blocks appended at the same offset.

**The control I care about:** disabling the `goalId` quarantine reds the conflict-resolved test itself. That is the proof that a hand-resolved test-file merge did not quietly become vacuous, which is the specific risk of resolving conflicts in test files, and the lane went looking for it without being told to.

On the substance, three corrections to my brief, all accepted. The one that changed the fix: **repairing by clearing the reference is structurally impossible for `activityTypeId`** — it is required on `TrackingEntry`, `Goal` and `RoutineBlock`, unlike `goalId?` and `routineBlockId?`, so there is nothing to clear it to. Quarantining *is* the repair, at every version. The outcome my issue actually wanted — a legacy store with a dangling activity type hydrates — is met, recoverably.

And it found a live second-order bug its own widened scope created: `entryIds` computed before the linkage phase would judge a pointer against the pre-linkage set and re-brick the launch for exactly the reason being fixed. Control C reds on it.

Two brick routes are left open **by decision and named**, with a test pinning the current strict behaviour so the gap is documented rather than latent — a goal or routine block with a missing `activityTypeId`, and a routine block with a dangling `goalId` at current version. Filed as #34, because setting a *skeleton* record aside needs a design decision no lane should take alone.

Carrying forward the sharpest unobserved item, which the lane raised itself: PR #32 and this PR are two independent fixes to the same durability path, merged without either author seeing the other's code. The suite says they compose; whether they do on a real torn write is behind #15.

Merging.

PR #37, opened 2026-09-12
fix(persistence): close legacy open timers at evidenced time, and say so (#4)

### LINKED ISSUE #4: F2: legacy migration silently discards the duration of a second open timer

RogueKostas:
Closes #4. **The last bug in the backlog.**

## The correction that matters most

**The existing test could not tell the two behaviours apart.** My brief warned about the
`repairLegacyValues` trap; there was a second instance of it sitting in the very test the brief told
the lane to replace. `tests/store/persistence.test.ts:113` asserted `openB.endTime === openB.startTime`
— but `makeTrackingEntry`'s default `updatedAt` is `10:00` while that fixture sets
`startTime: '11:00'`. `updatedAt` precedes `startTime`, so there is no evidence, and the *fixed* code
closes that entry at `startTime` too.

The lane wrote the fix, ran the suite, got 117/117 green, and noticed the assertion had never moved.
**A lane that trusted that test would have shipped nothing and called it done.**

## Two more corrections, both accepted

**"Never at `startTime`" is not achievable, and should not be.** When `updatedAt` is at or before
`startTime` and nothing follows, every later timestamp would be *invented*. The honest version
shipped instead: close at `startTime`, tag it `evidence: 'none'`, and report it — so an unevidenced
zero is never indistinguishable from a discarded one. That is the real fix to the word "silently" in
the issue title.

**My brief contradicted itself** — scope said `persistence.ts` and tests, deliverable 2 said surface
repairs *to the user*. A `getRepairedTrackingEntries()` that nothing renders is surfaced to a
programmer, not a user. The lane took the deliverable, touched `App.tsx`, `QuarantineNotice.tsx` and
two barrels, and flagged the expansion rather than burying it. My brief was wrong; the expansion is
correct.

**On "prefer reusing the quarantine sink" — right about the plumbing, wrong about the array.** A
quarantined record is *absent* from live state and preserved verbatim; a repaired one is *present*
and no longer verbatim. One shared array would make the shipped notice — "We couldn't read N tracking
entries, so they were set aside" — false about every repaired entry. The lane reused the record
shape, the `MigrationOptions` channel and #32's idempotent side-car, as a separate `repairs` channel
with its own sentence.

## The contract

`closeStrandedOpenEntry` (`src/store/persistence.ts:538`): close at `updatedAt` when it outlives
`startTime`; a later entry's `startTime` only ever **clamps that back**, because it proves the timer
was over, not that it ran that long; otherwise `startTime` with `evidence: 'none'`.

**Never destroy evidenced duration, never invent unevidenced duration.**

| Gate | Lane |
|---|---|
| `npm ci` from tracked lock | exit 0 |
| `npm test` | exit 0, **117 → 129** |
| `git diff --name-status origin/main...HEAD` | 9 `M`, no `D` |

Six negative controls, each RED then restored by exact inverse edit, with a final grep confirming no
mutation survived. The one that matters:
`expected '2026-03-02T11:00:00.000Z' to be '2026-03-02T12:30:00.000Z'` — **90 worked minutes
collapsed to zero.** Others: report disabled (6 red), clamp inverted (1 red — invented 7h instead of
1h), repair notice reusing the quarantine sentence (2 red), side-car repairs dropped (2 red),
selected-timer filter removed (7 red).

## It answered the composition question I attached — and found a real gap

I asked it to watch, while it was in that file, for any way PR #32 and PR #36 fail to compose. It
read `node_modules/zustand/esm/middleware.mjs` rather than the comments. The order is `migrate`
(:391) → `merge` (:415) → `set` (:419) → `if (migrated) setItem()` (:421).

**#32's awaited side-car write protects migrate-stage drops only.** A record dropped by the **merge**
stage during a *migrate-path* hydration is stripped from the blob at :421, while its side-car write
happens later in `initializeAppStore`, where failure is swallowed with a `console.warn`. The comment
there justifies swallowing on the grounds that ":405 short-circuits when the version matches" — true
of the pure merge path, **false for merge-stage drops on the migrate path**. That is exactly the #19
window #32 closed, still open on one branch.

Unreachable today, but only because every lenient repair happens to emit strict-valid output, and
**nothing enforces or tests that invariant**. The lane verified its own repair preserves it and did
not touch the gap — a durability change to merged work is not a lane's to make. Filed separately.

## NOT OBSERVED

- **#15 is open: no physical-device smoke has ever been run.** Only a device can settle whether blobs
  written by shipped v1/v2/v3 builds actually carry the `updatedAt` values this repair depends on —
  **the whole fix degrades to the old zero if they do not** — and whether the two banners stack
  legibly.
- **`updatedAt` is a proxy that can overstate.** `deleteGoal` and three routine-block paths stamp it
  for reasons unrelated to the timer running, so an open entry whose goal was deleted hours later
  closes at that deletion. Bounded by the clamp, strictly better than zero, still a guess — which is
  why the notice asks the user to check rather than asserting.
- A repaired `endTime` can land on a later calendar day than the entry's `date`. Pre-existing shape,
  not introduced here, not investigated.
- Entry selection unchanged: `openEntries.at(-1)` picks last-in-array, not latest-by-`startTime`,
  when the pointer names nothing open. Arguably wrong, out of scope, left alone.
- Lint was not run; it is not among this repo's gates and the lane did not check whether a script
  exists.


RogueKostas:
Orchestrator review, non-author. Gate run by me on the merged tree at `e5ad0fa` (0 commits behind origin/main): `npm ci` 0, `typecheck` 0, **129 tests**, `build:web` 0. Gated push clean, remote ref read back matching. **This is the last bug in the backlog.**

The finding I want on the record is the one about the test. My brief warned about the `repairLegacyValues` trap, and there was a second instance of it sitting inside the very assertion the brief told this lane to replace — a fixture whose `updatedAt` precedes its `startTime`, so the old behaviour and the new one produce the identical result. The lane wrote the fix, saw 117/117 green, and *noticed the assertion had never moved*. A lane that trusted its brief and its green suite would have shipped nothing and reported success.

I also accept that my brief contradicted itself — scope said persistence.ts and tests, deliverable 2 said surface repairs to the user, and a getter nothing renders is surfaced to a programmer. Taking the deliverable and flagging the expansion was the right resolution of my own inconsistency.

And "never at `startTime`" was my wording, not a requirement: when `updatedAt` is at or before `startTime` and nothing follows, every later timestamp is invented. Closing there but tagging `evidence: 'none'` and reporting it is the honest fix to the word "silently" — an unevidenced zero is no longer indistinguishable from a discarded one.

On reusing the quarantine sink: right about the plumbing, wrong about the array, and the reason is exact. A quarantined record is absent and verbatim; a repaired one is present and altered. One shared array would have made the shipped notice — "we couldn't read N entries, so they were set aside" — false about every repaired entry.

**The composition gap is a real find and it is now #38.** I attached that question as a side-task and it came back with an answer from the source rather than the comments: #32's awaited write protects migrate-stage drops only, and the comment justifying the swallowed merge-stage write gives a reason that is true of the pure merge path and false on the migrate path. Unreachable today only because every lenient repair happens to emit strict-valid output — and nothing enforces that. Leaving it untouched was correct; a durability change to already-merged work is not a lane's call.

Merging, and stopping the loop here.

PR #68, opened 2026-09-17
fix(web): replace every Alert.alert with an in-app dialog that works on web (#39)

### LINKED ISSUE #39: Alert.alert is a no-op on web: 19 flows silently do nothing on the only surface we can use
## Why

Found while preparing the review walkthrough, verified live against the deployed beta at `cdf3838`.

**`Alert.alert` is a no-op on React Native Web.** There are **23 call sites** in `src/`, 19 of them
outside the debug panel. On the web build every one produces nothing — no dialog, no console error,
no visual feedback. The control simply appears dead.

This matters more than a normal bug because **web is the only surface anyone can use.** `docs/PRODUCT.md`
D8 makes it a supported product surface, and #15 records that no build has ever run on a phone. So
these flows are not degraded on web, they are the only version of themselves that exists, and they
do nothing.

## Reproduce

Open https://zenroutine-web.onrender.com → Settings → **Privacy Policy**. Its handler is a single
`Alert.alert('Privacy', 'Your data stays on your device.')` with nothing else in it
(`src/screens/SettingsScreen.tsx:276`). Nothing happens. No console output.

## What is dead on web

| Screen | Flow |
|---|---|
| `SettingsScreen` | Theme picker, Load Sample Data, Reset All Data (confirm **and** the failure message), Export errors, Import errors, Privacy Policy |
| `ActivityTypesScreen` | Delete confirm, and both validation failures |
| `RoutineScreen` | "Replace existing blocks?" when copying a day |
| `BlockEditor` | "Time Conflict" on overlap, "Delete Block" confirm, every validation failure |
| `TrackingControls` | "Discard tracking?" and "Discard?" confirms |

Two consequences worth separating:

1. **Destructive actions silently do nothing** — a user clicks Delete and the item stays. Annoying,
   but safe.
2. **Validation failures are invisible** — a user saves a block that overlaps another, or a goal with
   a bad estimate, and gets no feedback at all. The form just sits there. That is the one that makes
   the app feel broken rather than incomplete.

## Deliver

1. A cross-platform confirm/alert. The options, roughly:
   - a small in-app modal component used everywhere (consistent with the app's own look, works on
     every platform, most work);
   - `window.confirm`/`window.alert` behind a platform shim (fastest, ugly on web, and cannot express
     the three-button Theme picker);
   - a library — but note `package-lock.json` is tracked and adding a dependency is a deliberate act,
     not a side effect.
   Pick one, say why in the code, and apply it at **every** call site rather than the loud ones.
2. The Theme picker is not a confirm at all — it is a three-option chooser built on `Alert.alert`
   with buttons. It needs a real control regardless of which option above is taken.
3. Validation failures should probably not be modal at all. Inline messages next to the offending
   field would be better on every platform; that is a design call, so raise it rather than assume it.

## Gates

in its own process. A negative control per new test. The suite is **129 at `cdf3838`** — re-measure at
your own tree and cite it.

**Also verify in a browser**, not only in vitest: `npm run build:web`, serve `dist/`, and click every
flow in the table above. A unit test that a component renders proves nothing about whether `Alert`
fires — that is exactly the gap that let this reach a deployed beta.

## Do not

`expo prebuild`, `eas build` or any Render command. Do not regenerate `package-lock.json` without
saying so explicitly and why.

## Done means

Every flow in the table does something visible on web, and a reviewer who clicks Delete either sees a
confirm or sees the thing deleted.

RogueKostas:
Closes #39.

Replaces all 25 `Alert.alert` call sites (23 in `src/`, 2 in `App.tsx`) with an in-app dialog that works on react-native-web, and adds a test that bans `Alert.alert` from coming back. Validation failures in the Block Editor and Activity Types are now shown inline.

Lane report (`web-dialogs`, headless claude, worktree):

# Lane web-dialogs (#39): every `Alert.alert` now opens an in-app dialog that works on web

Committed on `agent/web-dialogs` as `a23080e`; not pushed. Typecheck, the 148 tests and the web build all pass. Nothing has been clicked in a browser: this lane has no browser, so the click-through below is still to do.

## Corrections to the brief (flagged first)
1. **There were 25 call sites, not 23.** `App.tsx` has two more: the "Reset local data?" confirm on the load-error screen and its failure message. The brief's grep only covered `src/`. `App.tsx` is in scope, so I converted both, and the ban test scans `App.tsx` as well as `src/`.
2. **"Cannot Delete" could never appear.** `deleteActivityType` in `useAppStore.ts` doesn't throw when a type is in use; it logs a warning and returns. So the old error branch never ran, and deleting an in-use type would still have shown nothing. I left the store alone: after deleting, the screen checks whether the type still exists and, if so, says why it was kept.
3. **"Export errors" had no `Alert` call.** Export and import errors were already shown inline in the backup modal, so there was nothing to convert.
4. **One new file outside the listed scope:** `src/components/routine/blockEditorErrors.ts`, a small pure helper that maps each Block Editor validation error to its field. I added it so the mapping can be tested without a renderer.
5. **react-native-web detail that shaped the design:** a Modal attaches itself to the page when it mounts, even while hidden, and later ones draw on top. A dialog Modal that stayed mounted at the app root would appear *underneath* the Block Editor and the stop-tracking sheet. So the dialog's Modal is mounted only while a dialog is open, which keeps it on top.

No `docs/DESIGN-2019.md` exists on this base, so there was nothing in the design to conflict with.

## What changed, by file
- **`src/components/common/dialogQueue.ts`** (new, no React Native imports): the dialog queue and the `confirm` / `notify` / `choose` API.
  - Dialogs show one at a time, in order.
  - Each answer is tied to the dialog it came from, so a double click or a late Escape can't answer the next dialog.
  - Cancel, Escape and the backdrop all give `false` for a confirm and `null` for a choice.
- **`src/components/common/Dialog.tsx`** (new): `DialogProvider` and `useDialog()`, drawn in a React Native `Modal`.
  - On web, Escape cancels and focus is kept inside the dialog.
  - Keyboard focus starts on **Cancel**, so a stray Enter never confirms a destructive action.
  - The card is at most 400px wide with 16px margins; buttons wrap.
  - A comment explains why this approach was chosen over `window.confirm` or a library.
- **`src/components/common/index.ts`**: exports the above.
- **`App.tsx`**: mounts `DialogProvider` inside `ThemeProvider` and converts its own reset confirm.
- **`SettingsScreen.tsx`** (dialog calls only):
  - Theme is now a real three-option chooser that marks the current choice with ✓.
  - Load Sample Data and Reset All Data use `confirm`, followed by a success or failure message.
  - Import complete and Privacy Policy use `notify`.
- **`ActivityTypesScreen.tsx`**:
  - An empty name now shows an **inline** red message under the Name field, with a red border; it clears when you type.
  - Delete asks for confirmation, then shows "Cannot Delete" if the type was kept.
- **`BlockEditor.tsx`**:
  - Validation errors now appear **inline** under the Time or Activity Type section, and clear when that field changes.
  - Time Conflict stays a `notify`, as the brief says, since there's no single field to attach it to.
  - Delete uses `confirm`.
- **`RoutineScreen.tsx`**: "Replace existing blocks?" uses `confirm`. **Small behaviour change:** "Copied" now also appears after a confirmed replace, not only after copying to an empty day.
- **`TrackingControls.tsx`**: both discard prompts use `confirm`.
- **`DebugPanel.tsx`**: three error messages use `notify`; reset uses `confirm`.
- **Tests:**
  - `dialogQueue.test.ts` (14 tests): order, cancel, Escape, late or double answers, and the three API calls.
  - `blockEditorErrors.test.ts` (3 tests).
  - `noAlertAlert.test.ts` (2 tests): fails on an `Alert.alert(` call or an `Alert` import from react-native anywhere in `src/` or `App.tsx`, and also checks that the scanner itself works.

## Gate evidence
1. **`npm ci`**: exit 0 from the tracked `package-lock.json`, which I confirmed is tracked.
2. **`npm run typecheck`**: exit 0. No new `any`, `@ts-ignore` or `@ts-nocheck`; the old `error: any` in the Activity Types screen became `unknown`.
3. **`npm test`**: `Test Files 11 passed (11) · Tests 148 passed (148)`. The 8 files that already existed, run alone, give `Tests 129 passed (129)`, which matches the brief; the 19 new tests account for the rest.
4. **`npm run build:web`**: exit 0 (`Web Bundled 9237ms index.ts (711 modules)` … `Exported: dist`).
5. **Negative controls**: see below.
6. **`git diff --name-status origin/main...HEAD`**: 9 `M` and 5 `A` lines, no `D`. `git status` is clean.

## Negative controls
Each break was made with an edit, shown red, undone with the exact reverse edit, and shown green again. `git diff --stat` on the file was empty afterwards.

| # | Break | Red result |
|---|---|---|
| 1 | Put `Alert.alert('Error', …)` back into `DebugPanel.tsx` | `noAlertAlert` failed: `"src/components/debug/DebugPanel.tsx: Alert.alert call"` |
| 1b | Put `Alert` back into `RoutineScreen.tsx`'s react-native import | failed: `"src/screens/RoutineScreen.tsx: Alert imported from react-native"` |
| 2 | Removed the check that an answer belongs to the dialog on screen (`first.id !== id`) | `dialogQueue`: 1 failed (the double-press test) |
| 3 | Made Escape on a confirm resolve `true` | `dialogQueue`: 4 failed, including the Escape → `false` tests |
| 4 | Changed `'activityTypeId'` to `'activityType'` in the error mapping | `blockEditorErrors`: 1 failed (the message landed under `other` instead of `activity`) |

## Click-through for the orchestrator
Serve `dist/` and check each flow at about 400px and at 1920px wide. Every dialog should be a centred card over a dark backdrop, no wider than 400px. Escape or a click on the backdrop should close it as a cancel.

1. **Settings → Privacy Policy**
   - A "Privacy" card appears with "Your data stays on your device." and an OK button.
2. **Settings → Theme**
   - A "Theme" card lists System, Light and Dark, with the current one highlighted and marked ✓, plus Cancel.
   - Picking Dark switches the app to dark and the row now says "Dark".
   - Cancel or Escape changes nothing.
3. **Settings → Load Sample Data**
   - A confirm appears with Cancel and a blue **Load** button.
   - Cancel: nothing happens.
   - Load: a "Success — Sample data has been loaded." card appears, and the Activity Types count and Routine show the sample data.
4. **Settings → Reset All Data**
   - A confirm appears with a red **Reset** button.
   - Cancel: the data is kept.
   - Reset: a "Success — All data has been reset." card appears. If onboarding reappears behind it, the card should still be on top.
5. **Settings → Manage Activity Types**
   - **Delete a type that is in use** (after loading sample data): a confirm appears; after **Delete**, a "Cannot Delete — "<name>" is used by a goal, a routine block or a tracking entry, so it was kept." card appears and the row stays.
   - **Delete an unused type** (add one first): after the confirm, the row disappears.
   - **+ Add → clear the name → Save**: red text "Please enter a name for the activity type." appears under Name with a red border, the sheet stays open, and typing clears the message.
6. **Routine**
   - On a day that has blocks, under "Copy this day to…", pick another day that also has blocks.
   - A "Replace existing blocks?" card appears with a red **Replace** button.
   - Cancel: nothing changes.
   - Replace: a "Copied — Blocks copied to <Day>" card appears.
7. **Block Editor**
   - Open a block, or add one, and set its time to overlap another block, then Save.
     - A "Time Conflict" card should appear **on top of** the editor sheet; OK returns to the editor.
   - Set start time equal to end time, then Save.
     - Red "Block must have a duration greater than 0" text appears under Time.
   - Edit an existing block → **Delete Block**.
     - A confirm appears above the sheet.
     - Cancel: the sheet stays.
     - Delete: the sheet closes and the block is gone.
8. **Tracking → Stop Tracking → Discard**
   - A "Discard tracking?" confirm appears **above** the stop sheet.
   - Cancel returns to the sheet.
   - Discard closes both and the session is deleted.
9. **Keyboard (web)**
   - With any confirm open, focus starts on Cancel, Tab stays inside the dialog, and Enter on Cancel does nothing destructive.

## Not done / unverified
- **Not observed in a browser.** Everything above, especially the dialog appearing above the Block Editor and stop-tracking sheets and Escape routing, comes from reading react-native-web 0.21's Modal source, not from watching it.
- **iOS and Android have not been run.** On iOS, a dialog opened from the app root may not appear over the Block Editor's sheet; this doesn't affect web.
- **The load-error "Reset local data" flow in `App.tsx` was converted but not exercised**; it only shows when loading saved data fails.
- **The Activity Types empty-name check has no unit test.** It is a one-line check inside the screen, and testing it would mean rendering the screen, which this test setup can't do.
- **The Debug Panel's dialogs are converted but not click-tested.**


RogueKostas:
Landed by the orchestrator. Gate on the merged tree (f7b9a59): typecheck exit 0, **158 tests / 13 files**, build:web exported. Browser check on that build (1920px): Settings → Privacy Policy shows a dialog and Escape closes it; Load Sample Data → confirm → Load → "Success" dialog, and the data appears in Routine; Routine → copy Thursday onto Wednesday → "Replace existing blocks?" with a red Replace; Block Editor → Delete Block → the confirm renders **above** the editor sheet (the lane's main unverified claim, now observed). **NOT OBSERVED:** Theme chooser, Reset, Activity Types flows, inline validation, the tracking discard, the 400px layout of the dialog card. Noticed in passing, not from this PR: the Block Editor renders in light theme inside the dark app.

Issue #50, opened 2026-09-14
Activity type and estimate must be optional: the app has to work as a plain to-do list


RogueKostas:
From the design, p. 51, in ink:

> "**Activity Type and Estimation are optional** (this allows for using the app like a simple ToDo list if desired — i.e. no scheduling functionality)"

The mock shows a goal row whose activity type is a literal `?` box and whose estimate defaults to `1hr`.

Today both are effectively required to get anything useful, and an unlinked goal is a second-class citizen.

**Why it matters.** This is the on-ramp. Someone types three things they want to do, and the app is useful before they have built a routine at all. Everything else — scheduling, forecasting — becomes opt-in depth rather than a precondition.

**Fix.** Allow goals with no activity type and no estimate. Render `?` for the missing type and no forecast for the missing estimate — not an error, not a nag. Make sure the Goals screen, forecast engine and validation all tolerate them.

**Done when.** A goal can be created with a name alone, appears in the list, and nothing in the app errors or hides it.

---
*Raised from the 14 Sep 2026 review recording and the 2019 design doc. See `docs/REVIEW-2026-09-14.md`.*

Issue #45, opened 2026-09-14
Time picker and New Goal take over the whole screen for one field


RogueKostas:
**Observed (recording 13:35–14:10, 15:59–16:08).** Picking a block start time opens a full-screen wheel picker. Creating a goal opens a full-page form whose Save is in the far top-right corner.

> "this is a crazy widget — why is it not embedded? I should just have a little drop-down here."
> "now I have to go all the way to the top right. We don't need the whole page — I was expecting little pop-ups and things."

**Source.** `src/components/routine/TimePicker.tsx`; the New Goal route.

**Fix.** Inline/popover the time picker next to the field. Make New Goal a sheet or modal with the action adjacent to the form, not a full navigation push.

**Done when.** Neither a time change nor a goal creation requires a full-screen navigation, on web and on phone width.

---
*Raised from the 14 Sep 2026 review recording and the 2019 design doc. See `docs/REVIEW-2026-09-14.md`.*

RogueKostas:
Time-picker half landed in #72 (inline Start/End fields, no full-screen picker). The New Goal half is being replaced by the design's inline add field in Wave B (#50, #51); this issue closes with that PR.

PR #72, opened 2026-09-17
fix(routine): edit block times in place, no full-screen picker (#45)

RogueKostas:
Refs #45 (time-picker half; the New Goal half is covered by Wave B's inline add field, #50/#51).

Block start and end times are now two compact fields inside the editor. You can type a time or pick one from a 15-minute list, and the full-screen wheel picker is removed.

**Brief correction accepted:** an end time before the start time is an **overnight block** (the existing validation, Home and tests all rely on this), so it is not rejected. Its duration now reads e.g. `21h 45m (overnight)`.

Lane report (`inline-pickers`, headless claude, worktree):

I rebuilt the Block Editor's time section. Start and end are now two small fields inside the editor: you can type a time or pick one from a list that opens in place, and the full-screen wheel picker is gone. Typecheck, tests and the web build all pass, but I had no browser, so nothing has been seen on screen yet. It's committed on `agent/inline-pickers` as `df97b30`, not pushed.

## Corrections to the brief
1. **The brief's "end ≤ start must still be rejected" contradicts the code.** Today only end = start is rejected (`validation.ts:38`). End before start is saved as an **overnight block** (e.g. 23:00–07:00), which validation, Home (`HomeScreen.tsx:116`) and an existing test all rely on. I kept that behaviour; the duration now reads e.g. `8h (overnight)`. Your step "set end before start → inline error and Save refused" will **not** show an error. Setting end equal to start does. If overnight blocks should go, that's a product decision and needs `validation.ts`, which is outside my scope.
2. **The test count on this base is 158, not 129.** It is now 198.
3. **"Duration: 3h 0m" came from the old picker's hand-written formatting.** With `formatDuration` it reads `3h`.
4. **One file outside the listed scope:** a new file of pure helper functions, `src/components/routine/timeFields.ts`, so the behaviour can be unit-tested.
5. **Times still show as 24-hour `09:15`, like the rest of the app, not the design's `7am`.** The fields accept `7am`-style input but show it back as `07:00`.

## What changed, by file
- **`TimePicker.tsx`:** rewritten. The full-screen `TimePicker` and `TimePickerModal` are deleted, along with the hand-written duration. A new inline `TimeField` is a text box plus a ▼ button that opens a 15-minute list (00:00–23:45) in place. `TimeRangePicker` now lays out Start → End → Duration in one row that wraps.
  - Typing a valid time updates the block as you type.
  - An invalid time shows an error once you leave the field.
- **`index.ts`:** the exports for the two deleted components are removed. No `D` lines, since the file itself stays.
- **`BlockEditor.tsx`, time section only:** Save is refused, with a message under Time, while a typed time doesn't parse. The existing validation errors still show there. The helper that picks a new block's default times moved to `timeFields.ts` unchanged, so new blocks still default to one hour. "Link to Goal" is untouched.
- **`time.ts`:** `parseTimeOfDay` added in its own block at the end of the file. It accepts all the formats in the brief, plus `midnight`. It rejects `25:00`, `24:00`, `7:60`, `abc`, empty input, `13pm`, `0am` and `7:5`.
- **`timeFields.ts` (new):** the list of times, the nearest list entry for a typed time, the duration label, the typed-time error message, and the default-slot helper.
- **Tests (new):** `tests/core/parseTimeOfDay.test.ts` (29) and `tests/components/timeFields.test.ts` (11).

## Gate evidence
- `npm ci`: done.
- `npm run typecheck`: exit 0, no new `any`, `@ts-ignore` or `@ts-nocheck`.
- `npm test`: `Test Files 15 passed (15)` · `Tests 198 passed (198)` (was 158).
- `npm run build:web`: `Exported: dist`.
- `git diff --name-status origin/main...HEAD`: `M` BlockEditor, TimePicker, index, time.ts; `A` timeFields.ts and the two test files. No `D` lines.

## Negative controls
I made seven breaks, each with the exact inverse `Edit` afterwards:

| Break | Test that went RED |
|---|---|
| `12am` read as `12` | `12am`, `12:30am` |
| hour limit `23` → `99` | `25:00`, `24:00` rejections |
| minute limit `59` → `60` (run on its own) | `7:60` rejection |
| one extra list entry | "offers every 15 minutes…" |
| nearest-entry picks `i - 1` | nearest-entry test |
| overnight check `< startMinutes` → `< 0` | overnight label |
| hidden errors ignored by Save | "hides an error while … typing, but not from Save" |
| free-slot end set to the next block's start | "first free hour, one hour long" |

Results: RED run showed `Tests 9 failed | 31 passed (40)`, and the separate minutes run showed `1 failed | 28 passed`. After restoring, `git diff --exit-code` on `time.ts` and `timeFields.ts` was empty, `git status` was clean, and the full suite showed 198 passed.

## Click-through for the orchestrator (web, 400px and 1920px)
1. Routine → open a block. Under **Time** you should see a bordered box with: `Start` [HH:MM ▼] → `End` [HH:MM ▼], then `Duration` beside them. At 400px, Duration may drop to a second line. No separate "Select Time" screen should appear at any point.
2. Click Start's ▼. A list of about 5 visible rows opens **directly under the field**, scrolled near the current time. Pick `09:15`: the list closes and the field shows `09:15`.
3. Click in End, select its text, type `5pm`. Duration shows `7h 45m` (mid-typing, `5` alone shows an overnight duration). Click elsewhere and End shows `17:00`.
4. Save. The block card shows `09:15 - 17:00`.
5. Reopen it. Type `abc` into End and click elsewhere: `End: Enter a time like 7:15am or 19:30` appears in red under the box. Save does nothing and the message stays. Fix it to `17:00` and the message clears.
6. Set End to `09:15`, equal to Start, and Save: `Block must have a duration greater than 0` appears under Time and nothing is saved.
7. Set End to `07:00`, before Start: Duration shows `21h 45m (overnight)` and Save is **accepted** (see correction 1). Cancel if you don't want to keep it.
8. Click Start's ▼, then End's ▼: only one list is open at a time. Clicking into a text box closes any open list.
9. New block (+) on an empty day: `09:00` → `10:00`, `1h`. On a day with blocks, start and end are one hour apart.
10. Open and cancel a block after leaving an invalid time, then reopen it: the fields show the stored times, with no stale text or error.

## Not done / unverified
- Nothing was rendered, because I had no browser. These are unverified:
  - that the list scrolls to the current time when it opens;
  - that clicking Save on web runs after the field has finished committing (Save still refuses an invalid time either way);
  - how the row wraps at 400px.
- The list is a scrollable list drawn by the app, not a native `<select>`, because React Native has no select control and no dependency was added.
- The editor itself is still the existing page-sheet modal, not the design's small dialog. That wasn't in this brief.


RogueKostas:
Landed by the orchestrator. Gate on the merged tree (9b58ccc485a75f11aeea456b726c97e275e72bbc): typecheck exit 0, **261 tests / 19 files**, build:web exported. Browser check at 400px on that build: opening a routine block shows **Start / End text fields with ▼ list buttons inline in the editor**, and there is no "Select Time" screen; typing `5pm` into End (start 09:00) updates Duration to `8h`. **NOT OBSERVED:** the dropdown list, the invalid-time error and Save refusal on screen (covered by unit tests), 1920px, overnight label. Remaining #45 half (New Goal as a full page) goes to Wave B (#50/#51); leaving #45 open until then.

PR #75, opened 2026-09-17
feat(engine): fill-forward forecast scheduler (#52, part 1)

RogueKostas:
Refs #52 (part 1 of 2: the engine; the calendar UI follows).

Adds `src/core/engine/forecast.ts`: a pure fill-forward scheduler that takes the routine and goals in list order and returns each goal's completion date, % milestones, and per-block allocation slices. It needs no tracking history, never reads `goal.priority` or `RoutineBlock.goalId` (see #49 and #60), and no screen uses it yet.

The director's example from the 14 Sep review (27:18) is a test and passes slice by slice.

Lane report (`forecast-engine`, headless claude, worktree):

I built the forecast engine as a new pure module. The director's example from review 27:18 passes exactly as he described it, and all four gates are green.

## Algorithm in five lines (for briefing the calendar and Home lanes)
1. Walk calendar days forward from `from`: first the day before (for the after-midnight part of an overnight block), then up to `horizonDays` days.
2. Take each day's blocks in start-time order. Only block time after `from` counts, and a partly elapsed minute counts as gone.
3. All of an activity type's block minutes go to that type's goals, strictly in the order of the goals list. `goalId` and `priority` are never read.
4. A goal's remaining minutes (default: estimate minus logged, never below 0) fill slices until they reach 0. That minute is the goal's completion; a % milestone is the minute its share of the whole estimate is reached, counting time already logged.
5. An overnight block belongs to the day it starts, so its minutes run past 1440 on that day's scale. A goal that is still open at the end of the horizon, or whose type has no blocks, is `no-capacity`.

## Corrections to the brief
- **Base and test count:** this branch is based on `dc8e276`, not `41a90cc`. The suite has **261** tests on this base, not 129; it is now **282**.
- **New `'inactive'` reason.** The brief's four reasons have no place for `paused` or `archived` goals, so I added `'inactive'` for them. `'completed'` is kept for completed goals only.
- **`no-capacity` goals keep their partial slices.** A goal still open at the end of the horizon keeps the slices and milestones it did get, and is also listed in `unscheduled`.
- **Overnight blocks** go further than the brief asked: an allocation's `endMinutes` can exceed 1440. The completion's `date` is the day the block starts, while `dateTime` gives the real clock time (e.g. Monday's forecast day, `2026-09-22T01:00`).
- **Yesterday's overnight block:** if `from` falls in its after-midnight part, that remaining time counts, and the slice is dated yesterday.

No design conflicts, and no files outside the scope were touched.

## What changed
[...]

Issue #49, opened 2026-09-14
Goal priority should be list order, not a five-value enum


RogueKostas:
The design makes priority **ordinal position in the goals list**, and says so in ink on p. 60 and p. 65:

> "**Drag and move rows around to reprioritise**"

He restated it unprompted at 26:17:
> "we're moving things up and down — just grabbing a goal and moving it higher up the list or lower. And that changes the priority. **We're not even showing the priority here; there's no column for priority, we're just ordering goals.**"

Today the app has a `Very High / High / Medium / Low / Very Low` chip group on the New Goal form and a priority badge on the card. Ties are then broken arbitrarily, which the forecast engine has to resolve somehow.

**Why it matters beyond taste.** The forecast's fill-forward sequencing (see the forecast-calendar issue) needs a *total order* over the goals in an activity type. An enum gives a partial order with ties; a list gives exactly what the algorithm needs. It is also the handle an AI assistant would manipulate — "push that right to the top" (28:48).

**Fix.** Add an explicit `order` (or fractional rank) per goal, scoped per activity type. Drag to reorder within a filtered list. Retire the enum, migrating existing values to a starting order.

**Done when.** Reordering persists, the forecast consumes the order, and no priority chips remain.

---
*Raised from the 14 Sep 2026 review recording and the 2019 design doc. See `docs/REVIEW-2026-09-14.md`.*

PR #78, opened 2026-09-17
feat(goals): priority is list order; forecast follows it (#49, model half)

RogueKostas:
Refs #49. This PR does the model half; the drag-to-reorder UI follows with #50/#51, and #49 closes then.

Goal priority is now the goal's position in one list of all goals, as the design and the director's settled decision require. The priority enum, chips and badge are gone. Schema v7 converts old priorities into a starting order (highest first, then oldest, then id). `moveGoal(goalId, { before | after })` is ready for the drag UI, and the forecast now takes its dates from the fill-forward engine (#75) in list order.

Lane report (`goal-order`, headless claude, worktree):

The priority enum is gone: goal priority is now the goal's place in one list of all goals, and the forecast uses that order. All gates pass (typecheck, 367 tests, the same under `TZ=UTC`, `build:web`), and it's committed on `agent/goal-order` as `f1c632f` plus `71194df`. Nothing was looked at in a browser; the click-through list is at the end.

## Corrections to the brief
- **Test count:** the base (`a30a862`) has **329** tests, not 304. The branch has **367**.
- **Two small changes outside the listed files:**
  - New file `src/core/engine/goalOrder.ts`, holding the pure ordering functions. It's new, so it can't collide with another lane.
  - `DebugPanel.tsx`: one label renamed from "Weekly allocation:" to "Next 7 days:", because the value behind it changed meaning (see Forecast).
- **The design agrees with "one list, not one per type".** p60 shows one list mixing types, and p65–67 reorder inside a filter. The design never shows how a filtered reorder merges back; I used your reading (move relative to the visible neighbours).
- **Priority changes never touched forecast confidence.** Confidence is reset only by routine block edits and routine activation (#7, #23, #24), and editing a goal's priority never did that. So a reorder resets nothing either, and a test checks that routines are untouched.
- **`src/components/goals/GoalCard.tsx` is never rendered.** It still labels a value "Weekly allocation"; I left it alone.

## What changed
- **Model (`Goal.ts`):** `priority` and its three helpers (`GoalPriority`, `PRIORITY_LABELS`, `PRIORITY_COLORS`) are removed. Goals now have an integer `order` numbered 0, 1, 2… with no gaps, and the stored goals list is always kept in that order.
  - I chose a plain integer over a fractional rank. The list is tens of goals on one device with no second writer, so renumbering on a move costs nothing. Fractional ranks would need rebalancing when they run out of precision.
  - New goals go to the bottom.
- **Migration, schema 6 → 7 (`persistence.ts`):**
  - The step has its own version gate (`GOAL_ORDER_SCHEMA_VERSION = 7`); `STRICT_SCHEMA_VERSION` stays 4.
  - Starting order is highest priority first, then oldest `createdAt`, then `id`. Old values are validated exactly as before.
  - At v7, a goal must have a whole, non-negative `order`, and two goals sharing a position is refused. Gaps are closed without changing relative order.
  - Old backups go through the same path, so they import with the same order.
- **Store (`useAppStore.ts`):**
  - `moveGoal(goalId, { before | after: goalId })`. I named the target by a neighbour rather than an index because a filtered view doesn't know where the hidden goals sit. Hidden goals keep their order relative to each other.
  - Moving a goal onto itself, to where it already is, or next to an unknown goal writes nothing.
  - `addGoal` appends and no longer stores a stray `priority` from an old caller.
  - `deleteGoal` closes the gap; `updateGoal` can't change `order`.
  - Example data is added below the user's own goals.
  - **Selector:** `useGoals` already returns list order, so no new hook was needed.
- **Forecast (`prediction.ts`):**
  - `predictAllGoals` now takes its dates from `forecastGoals`, in list order. The confidence logic is unchanged.
  - `PredictionResult` keeps its shape plus one new field, `goalsAhead`.
  - `weeklyMinutesAllocated` now means the minutes the goal actually gets in the next 7 days.
  - The completion date is the day the last block starts. For a block that runs past midnight, that's the day before the clock time.
  - Tests now pin a local fake clock.
- **Screens and copy:**
  - **`GoalsScreen.tsx`:** the priority chips, the card badge and the priority sort are removed. The explainer text is updated, and the per-goal line now reads e.g. "Next in line after N goals of this type."
  - **`blockGoals.ts`:** the Block Editor list is in list order.
  - **Onboarding:** slides 2 and 3 no longer mention priorities.
  - **Sample data:** ordered dashboard, marathon, course, planning doc, reading. The confident/low-confidence mix still holds.
- **Tests:**
  - `tests/fixtures/v6-priority-store.json` was written by the unchanged store's own actions before any edit, with pinned clock and ids. It includes a same-millisecond tie, a completed goal, a paused goal, an overnight block and a running timer.
  - New files: `goalOrderMigration.test.ts`, `goalOrder.test.ts`, and `invariants/noGoalPriority.test.ts` (a type-level check plus a scan of `src/`, allowing only the migration file).
  - The director's example now also runs through the prediction path.
  - Older fixture tests now expect the derived order.

## Gate evidence
- `npm ci` exit 0; `package-lock.json` is tracked and unchanged.
- `npm run typecheck`: exit 0, no new `any` or suppressions.
- `npm test`: 26 files, 367 passed.
- `TZ=UTC npm test`: 367 passed; the start time printed in UTC, so the zone took effect.
- `npm run build:web`: exit 0, "Exported: dist".
- `git diff --name-status origin/main...HEAD`: 5 added, 18 modified, no deletions.

## Negative controls
Each break went red, was restored with the exact inverse edit, went green, and left `git diff --stat -- src/` empty.

| Break | Went red |
|---|---|
| migration without the `id` tie-break | 4 migration tests, including the fixture and hydration ones |
| shared-position check disabled | "refuses a shared order" |
| negative-order check loosened | "refuses a negative order" |
| v7 read without sorting | "sorts by order…", "ignores a priority left on a v7 goal" |
| export drops `order` | export/re-import test, backup round-trip |
| schema gate moved to 8 | 11 migration tests |
| "after" inserted as "before" | 4 move tests |
| every goal's `updatedAt` stamped on a move | "stamps only the goal that was moved" |
| unstable sort, delete without renumbering, append at 0 | their 5 tests |
| unknown-target check removed | red only after I strengthened the test |
| `updateGoal` order guard removed | red only after I strengthened the test |
| stray `priority` kept, routines touched on move, example goals not renumbered | their 3 tests |
| forecast queue reversed | 6 prediction tests, including the director's example |
| forecast ignores `order` | 2 prediction tests |
| `goalsAhead` forced to 0, 7-day window off by a day, plural copy, "priorities" in the explainer | 10 tests |
| Block Editor list unsorted, old onboarding copy, `priority?` added back to `Goal` | their tests; `tsc` failed on the last |

The two controls that first stayed green caught weak tests:
- **Unknown target:** my test only moved the top goal, where the broken code happens to rebuild the same list.
- **`updateGoal` guard:** it only matters when a status is passed too.

Both tests were fixed in `71194df`. The "fixture really is v6" test checks data only, so there is no production code to break.

## Click-through for the orchestrator (web, ~400px and 1920px)
1. **Reload over an existing v6 store:** every goal is still there. Order is highest old priority first; within a priority, oldest first. Nothing is flagged as set aside, and the stored blob now has `"version":7` and `"order"` on each goal, with no `"priority"`.
2. **Goals → + Add:** New Goal shows Name, Description, Estimated Time and Activity Type only, with no Priority chips.
3. **Goal cards:** only the type name under the goal name, with no coloured priority badge.
4. **Add a goal:** it appears at the bottom of the "All" list.
5. **"How forecasts work":** reads "…That type's active goals are worked in list order: the top one gets all of the time first, and the next one starts when it is done. Dates assume this routine and this order continue."
6. **Per-goal forecast:**
   - First line: "Xh in the next 7 days, of Yh/week scheduled for this type".
   - Second line: "First in line: this type's time goes here first.", "Next in line after N goal(s) of this type." or "The only active goal of this type.", followed by the confidence reason.
7. **Routine → edit a block:** "Goals for this activity type" matches the order of that type's goals on the Goals screen.
8. **Onboarding slide 2** reads "List your goals with time estimates, most important first. The top of the list is worked on first."; **slide 3** ends "Your goals take that time in list order."
9. **With example data:** the Work goals are "Ship the analytics dashboard" (first in line) and then "Write the Q4 planning doc" (next in line after 1 goal).

## Not done / unverified
- Nothing was checked in a browser.
- There's no drag UI; that's the `goals-list` lane (#50/#51).
- I didn't test other time zones: `TZ` settings other than UTC seemed to have no effect on this Windows machine, so I didn't count those runs.
- Forecasts only look ahead as far as the remaining work needs, capped at about 100 years. A goal that can't finish within that shows "No available routine time".


RogueKostas:
Landed by the orchestrator. CI green (read before merging). Gate on the merged tree (8f03892): typecheck exit 0, **395 tests / 27 files** under the host zone and TZ=UTC, build:web exported. Browser check: a **real v6 store** with priorities Ship:1, Q4:3, Marathon:2, TS:2, DeepWork:4, Test:3 reloaded as **v7** in the order 0 Ship, 1 Marathon, 2 TS, 3 Q4, 4 Test goal, 5 Deep Work, with no `priority` key left. Goals explainer reads "…worked in list order: the top one gets all of the time first…". Per-goal lines read "First in line…" and "Next in line after 1/2 goals of this type". The New Goal form has no priority chips. **NOT OBSERVED:** onboarding copy, 400px, and a reorder (no UI yet: #50/#51).

PR #69, opened 2026-09-17
fix(onboarding): five working slides, and example data from the first screen (#40, #62)

RogueKostas:
Closes #40. Closes #62.

Onboarding shows five distinct slides on web, with Next and Back. "Try it with example data" appears on the last slide and on an empty Home. The example data is rebuilt so that forecasts show both high and low confidence.

**For the director (flagged by the lane, not changed):** onboarding slide 3 says "Link activities to goals", and slide 2 mentions "priorities". Both contradict settled decisions (#60, #49). The Wave B lanes own that copy.

Lane report (`first-run`, headless claude, worktree):

Onboarding now shows five distinct slides with working Next and Back, and "Try it with example data" is on the last slide and on an empty Home. All gates pass: 129 → 143 tests, and every new test went red under a deliberate break. Nothing has been checked in a browser, because I don't have one. The work is committed on `agent/first-run` as `3519017`; nothing was pushed and no PR was opened.

## Corrections to the brief (read first)
1. **The old sample data could never show a confident forecast.** A forecast only counts tracking done after the routine's schedule last changed. The old code built its routine through `addRoutineBlock`, which records that change as "now", so any backdated history was ignored. The example routine is now dated to before its history.
2. **The old sample data could hide a user's real routine.** Loading it from Settings made a new routine active, and the app has no way to switch routines, so the user's own schedule disappeared from view. It could also do nothing at all if a default activity type had been renamed.
3. **After Reset All Data, the app shows onboarding first, not Home.** Reset also clears the "onboarding done" flag. So the last click-through step is: reset, reload, Skip, then Home shows the offer.
4. **There is no "offset" function.** The brief listed one, but the screen now renders only the current slide, so nothing scrolls.
5. **Onboarding copy conflicts with settled decisions (flagged, not changed).** Slide 3 says "Link activities to goals", which contradicts "the routine is made of activity types only". Slide 2 mentions "priorities", but priority is becoming list order.
6. **Settings shows "Sample data has been loaded." on a second tap**, although nothing loads the second time. `_addSampleData` now returns false in that case, but the brief didn't allow edits to `SettingsScreen.tsx`.
7. **Nothing outside Scope was touched.** `docs/DESIGN-2019.md` isn't on this base yet.

## What changed, by file
- **`src/screens/onboardingPaging.ts` (new):** the five slides plus the paging logic as plain functions: index clamping, Next/Back, button label, dots, and slide width from the measured container.
[...]

Issue #54, opened 2026-09-14
Scheduled, started and tracked are three different states — and untracked time must be visible


RogueKostas:
From the design, p. 77, in ink:

> "**By default activities 'start' when scheduled but do not 'track'.** The user has to actively indicate that the app can start tracking. The user can also pause tracking. In the use case below the user started the activity some time after it was scheduled and later had to pause."

And, labelling the grey wedges of the pie:
> "**Time spent not tracking**"

So a block moves through: *scheduled* → *started* (the clock says it's happening) → *tracking* (the user confirmed) → *paused* → *ended*. The shortfall between scheduled and tracked is shown as grey in the timer pie — adherence rendered as a visible gap rather than a score.

Today tracking is binary and the gap is invisible.

**Relationship to #13.** This is the live half of that issue. The retroactive half is the catch-up queue. Both feed the same adherence model but with different confidence, which is exactly what #13 asks for.

**Done when.** The four states are modelled and persisted, the timer pie renders scheduled-but-untracked time in grey, and pausing is a first-class action.

---
*Raised from the 14 Sep 2026 review recording and the 2019 design doc. See `docs/REVIEW-2026-09-14.md`.*

PR #85, opened 2026-09-17
fix(store): close the merge-stage side-car window, and test that repairs stay strict-valid (#38)

### LINKED ISSUE #38: Merge-stage side-car write is still swallowed on the migrate path, and the invariant that hides it is untested
## Why

Found by the `open-timer-duration` lane (#4) while working in `src/store/persistence.ts`. It was
asked to watch for any way PR #32 and PR #36 fail to compose, since those two independent durability
fixes were merged without either author seeing the other's code. It found one, by reading
`node_modules/zustand/esm/middleware.mjs` rather than the comments around it.

## The gap

zustand's order on a hydration is:

```
migrate (:391) -> merge (:415) -> set (:419) -> if (migrated) setItem() (:421)
```

**PR #32's awaited side-car write protects migrate-stage drops only.** A record dropped by the
**merge** stage during a *migrate-path* hydration is stripped from the app blob at `:421`, while its
side-car write happens later, in `initializeAppStore`, where a failure is **swallowed** with a
`console.warn`.

The comment justifying that swallow says `:405` short-circuits when the version matches. That is true
of the pure merge path. It is **false for merge-stage drops on the migrate path**, and that is
precisely the window #19 measured and #32 closed — still open on one branch.

If the side-car write fails there, the record is gone from both keys and hydration still reports
success. Same failure mode as #18, reached by a different route.

## Why it is not on fire

**It is unreachable today** — but only by accident. Every lenient repair currently happens to emit
strict-valid output, so the merge stage has nothing to drop on a migrate-path hydration. **Nothing
enforces or tests that invariant.** The #4 lane verified its own repair preserves it (all three
evidence branches yield `endTime >= startTime`, exactly one entry stays open) and left the gap
alone, correctly: a durability change to already-merged work is not a lane's call.

The next lenient repair that emits something the strict pass rejects reopens it, and nothing will say
so.

## Deliver

1. **Make the invariant explicit and tested**, rather than true by luck: every lenient repair path
   must emit output the strict pass accepts. A test that walks the repair branches and asserts strict
   acceptance is the cheap version.
2. **Then decide the swallow.** Either close the window — the merge-stage side-car write becomes
   awaited or reordered the way `migrate`'s was in #32 — or record in the comment exactly why
   swallowing is safe, with the invariant above cited as the reason. The current comment gives a
   reason that is not true in general, which is worse than no comment.
3. Correct that comment either way.

## Gates

in its own process. A negative control per new test: RED before, GREEN after, both pasted. The suite
is **129 at the tip of PR #37** — re-measure at your own tree and cite it.

## Do not

## Done means

The invariant that keeps this unreachable is asserted by a test that fails if it stops holding, and
the swallowed write is either fixed or justified by a reason that is actually true.

RogueKostas:
Closes #38.

The strict read now happens inside `migrate`, via a new `hydratePersistedState`, so every record either read sets aside reaches the awaited side-car write **before** zustand rewrites the app blob. The merge-stage window on the migrate path is closed structurally, not documented away, and the comment that justified swallowing the failure (which was only true of the pure merge path) is rewritten.

The "every lenient repair emits strict-valid output" invariant is now tested: a 22-case table walks every repair and migration branch, asserts the repair actually fired, and then asserts the strict read accepts the result unchanged.

**Two corrections the lane made to my brief and to the code:**
- The trap is narrower than the issue says. The call is `parseTrackingEntry(value, version < STRICT_SCHEMA_VERSION, version)`, so only blobs **below v4** are read leniently; v4–v8 are strict. The same wrong constant was written into a `useAppStore.ts` comment and is fixed.
- `readTrackingPauses` (v9, #54) is the only read that is looser below CURRENT than at CURRENT. It stays safe because it drops the key unread — and there is now a test that fails if that gate is removed.

**How the fix is proven:** negative control NC1b reproduced #38 exactly. With a repair deliberately emitting strict-invalid output and the side-car key unwritable, the pre-fix code reported hydration `ready` with the record in **neither** key and only a `console.warn`; with the fix, hydration reports `error` and the record stays in the blob. There is no positive test, because while the invariant holds no input reaches the window; the lane says so plainly rather than inventing one.

Lane report (`sidecar-merge`, headless claude, worktree):

All gates green. Here is the report.

---

# Lane report — sidecar-merge (#38)

## 1. Corrections to the brief

**The brief (and a code comment) name the wrong constant, and it changes the trap's scope.**
The brief's Pass 9 trap says `parseTrackingEntry(value, version < CURRENT_SCHEMA_VERSION)`. On my base the signature is `parseTrackingEntry(value, repairLegacyValues, version)` and the call is `parseTrackingEntry(value, version < STRICT_SCHEMA_VERSION, version)` — `persistence.ts:838`. `STRICT_SCHEMA_VERSION` is 4. So a **v4–v8** blob on the migrate path is parsed *strictly*, not leniently: its malformations are quarantined, not repaired. Only a blob below v4 is lenient. The trap is real but narrower than stated, and I built every lenient case at v3 and asserted the repair fired.

The same wrong constant was written into production: `useAppStore.ts` merge stage said "the `version < CURRENT_SCHEMA_VERSION` branch that repairs open entries cannot run here". Corrected to `STRICT_SCHEMA_VERSION`. That comment is the likely source of the brief's wording.

**"Prefer closing it" was achievable — the alternative wasn't needed.** Deliverable 2's fallback ("document why swallowing is safe") is not used: the window is closed structurally and the comment now says so.

**The issue's stale numbers.** Issue #38 says "the suite is 129 at the tip of PR #37". Base is **697 at 62c522e**, matching the lane brief.

**zustand line numbers re-read on this tree and confirmed** — `node_modules/zustand/esm/middleware.mjs`: `:390` version check, `:392` migrate call, `:405` short-circuit, `:415` merge, `:419` set, `:421` setItem, `:431` catch. The issue's `:391` is the `if (options.migrate)`; the call is `:392`. Everything else is exact.

**One addition to the issue's account of *why* it is unreachable.** The issue says it's because every lenient repair emits strict-valid output. I walked all of it and that is right, but there's a second mechanism worth naming: `readTrackingPauses` (v9, #54) is the **only** read in the file that is *looser* below CURRENT than at CURRENT. It stays safe because it drops the `pauses` key unread rather than accepting a bad value — so its output is strict-valid too. That is the branch the issue could not have anticipated, and it's the one my brief told me to make sure I covered. If `readTrackingPauses` ever stopped gating on the version, the window would open. There is now a test for exactly that.

**Scope note (flagged as required).** I also corrected the `MAX_QUARANTINE_GENERATIONS` doc comment in `persistence.ts` — a correction `useAppStore.ts:183` records as "owed there" ("realistically unreachable" was wrong when written). In-file scope, outside the brief's deliverables, comment-only.

## 2. What changed, by file

**`src/store/persistence.ts`**
- New exported `hydratePersistedState(persistedState, version, options?)` — the versioned read, then the strict read of its result, **against one set of sinks**. This is the fix: the strict read now happens somewhere that can be awaited.
- `MAX_QUARANTINE_GENERATIONS` comment corrected.

**`src/store/useAppStore.ts`** (hydration + `initializeAppStore` only)
- `migrate` calls `hydratePersistedState` instead of `migratePersistedState`. Every record either read drops is therefore on the awaited `persistHydrationReports` call, **before** zustand's `set` (`:419`) and `setItem` (`:421`) rewrite the app blob.
- `merge` unchanged in behaviour. Its comment now states what the call is on each path, and that on the migrate path it is a re-read of a state that already passed the strict read.
- `initializeAppStore`'s swallow comment rewritten. The old reason (":405 short-circuits when the version matches") was true only of the pure merge path; the new one enumerates both paths, states what was false, and says why the swallow is now safe by construction.
- Shared-sink comment and the `STRICT_SCHEMA_VERSION` correction.

**`tests/store/repairsAreStrictValid.test.ts`** (new, 28 tests)
- A 22-case table walking every repair and migration branch: goal estimate, `completedAt` both directions, `loggedMinutes`, v6→v7 priority→order, order gaps, v7→v8 optional goal fields, v5→v6 block `goalId`, v1 icon names, `endTime < startTime`, stale entry `goalId` and `routineBlockId`, `activeRoutineId`, **all three** open-timer evidence branches, the adopted timer pointer, `hasCompletedOnboarding`, `preferences` and `capacityChangedAt` fallbacks, the v8→v9 pauses gate, and one case with all of them at once.
- Each case asserts **the repair fired** (`expectRepaired`, against the versioned read's own output — the anti-vacuity guard for the trap), then that the strict read accepts the result **unchanged** (`toEqual`, not merely "didn't throw"), with an empty quarantine, and also that a sink-less strict read doesn't throw.
- A table-integrity test: at least one case below the repair gate, at least one above it, and none at CURRENT.
- Strict-read idempotence (2 tests) — the other half of what makes `merge`'s re-read a no-op.
- `hydratePersistedState` (2 tests): it reports the versioned read's drops to the caller, and it adds nothing of its own to either the state or the sink.

## 3. Gate evidence

| Gate | Result |
|---|---|
| `npm ci` | `added 584 packages, and audited 585 packages in 16s` — exit 0. Lockfile tracked (`git ls-files --error-unmatch package-lock.json` → `package-lock.json`) and untouched. |
| `npm run typecheck` | exit 0, no output. No new `any` / `@ts-ignore` / `@ts-nocheck` (`tests/invariants/noTypeCheckSuppressions.test.ts` green). |
| base suite | **697 passed (43 files)** at 62c522e — matches the brief. |
| `npm test` | **725 passed (44 files)**, exit 0. 697 + 28. |
| `TZ=UTC npm test` | **725 passed (44 files)**, exit 0. |
| `npm run build:web` | `Exported: dist`, exit 0. |
| `git diff --name-status origin/main...HEAD` | `M src/store/persistence.ts` · `M src/store/useAppStore.ts` · `A tests/store/repairsAreStrictValid.test.ts` — no `D` lines. |

Commits: `6b435e4` (fix + tests), `40620a8` (two test assertions the negative controls showed were weak).

## 4. Negative controls

Committed first, so every restore is proved by an empty `git diff` on the production files. All six restored; final `git status` on `src/` is clean.

**NC1 — a lenient repair emits strict-invalid output.** `persistence.ts:711`, `endTime = startTime` → `endTime = new Date(Date.parse(startTime) - 1).toISOString()`. This is precisely the scenario the issue predicts.

```
× an endTime before its startTime is pulled back to the startTime
× every legacy repair at once still composes
× a store with a repaired legacy history
× adds nothing of its own to the state or to the sink on a legacy blob
Tests  4 failed | 24 passed (28)
```

**NC1b — the same break, used to demonstrate the window itself.** With NC1 in place I ran a throwaway store-level probe (`tests/store/probe38.test.ts`, deleted after; `rm` succeeded, `git status` empty): a v3 blob holding the backwards entry, with the side-car key made unwritable.

With the fix:
```json
{ "hydration": { "status": "error", "error": "disk full" },
  "inBlob": true, "inSidecar": false, "liveEntries": [] }
```
Then, keeping the broken repair, I reverted only the fix (`migrate` back to `migratePersistedState`):
```
stderr: ZenRoutine could not save its hydration report Error: disk full
        at persistHydrationReports (src/store/useAppStore.ts:211:22)
{ "hydration": { "status": "ready", "error": null },
  "inBlob": false, "inSidecar": false, "liveEntries": [] }
```
**`ready`, and the record in neither key**, with a `console.warn` as its only trace. That is issue #38 reproduced, and the fix closing it.

**NC2 — remove the v9 pauses version gate** (`if (version < TRACKING_PAUSES_SCHEMA_VERSION) return undefined;` deleted):
```
× a pre-v9 entry's pauses are dropped unread, malformed or not (#54)
AssertionError: expected [ { index: +0, …(3) } ] to deeply equal []
Tests  1 failed | 27 passed (28)
```

**NC3 — break a repair in a way no per-case assertion can see.** `orderGoalsByLegacyPriority`'s `.map(({goal}, order) => ({...goal, order}))` → `order: 0`. The combined v1 case has no assertion on goal order at all, and it went red on the *strictness* assertion:
```
× a pre-v7 priority enum becomes a list order (#49)
× every legacy repair at once still composes
Error: Invalid goals: two goals share a list position
Tests  2 failed | 26 passed (28)
```
This is the control that matters most: it proves the generalising half of each test is live, not just the hand-pinned expected values.

**NC4 — make the strict read non-idempotent.** `notes: readOptionalString(record,'notes')` → `` `${… ?? ''}.` ``. Both idempotence tests red, plus ten others:
```
× a store with history
× a store with a repaired legacy history
… Tests  12 failed | 16 passed (28)
```

**NC5 — drop the sink from the first read inside `hydratePersistedState`.** Red in the new file *and* in the existing durability tests, confirming the new function is really on the hydration path:
```
× reports the versioned read's drops to the caller
× adds nothing of its own to the state or to the sink on a legacy blob
× writes the side-car before the migrate path rewrites the app blob
× leaves the record in the app blob when the side-car write fails on a pre-v4 blob
… Tests  7 failed | 49 passed (56)
```

**NC6 — swallow failures in `hydratePersistedState`:**
```
× propagates a read that throws rather than swallowing it
AssertionError: expected [Function] to throw an error
Tests  1 failed | 27 passed (28)
```

NC6 also **found a weak test**: a `try/catch` around only the *second* read left all 28 green. The strict read cannot be made to throw while the invariant holds, so that test can only ever exercise the versioned read. I did not delete it — I corrected its comment to say which read it covers. Commit `40620a8`. I removed a third test in the same pass (`does not read a current-version blob twice`) because it was unfalsifiable: with the short-circuit removed it stayed green, since the second read is idempotent. An assertion that cannot fail is not a test.

## 5. Click-through for the orchestrator

**There is none, and I want to be blunt about why:** this lane changes no UI, no screen, no component and no user-visible behaviour. Nothing renders differently at 400px or 1920px. A normal click-through would confirm only that the app still opens, which the 725-test suite and `build:web` already say.

If you want to observe the change anyway, it has to be done through DevTools, and I have measured the two facts that make the instructions accurate:
- on web, `@react-native-async-storage/async-storage` is `window.localStorage` **with the key verbatim** — `zenroutine-storage` and `zenroutine-quarantine`, no prefix (`lib/module/AsyncStorage.js:59-65`);
- the persist format is `{"state":{…},"version":N}` (`createJSONStorage`, `useAppStore.ts:635`).

**Observation A — the migrate path still works end to end (2 min).**
1. Serve the lane's `dist` on `:8765` and open it.
2. DevTools → Console. Read the current blob: `JSON.parse(localStorage['zenroutine-storage'])`. Note `version` is `9`.
3. Re-stamp it as legacy and reload:
   ```js
   const b = JSON.parse(localStorage['zenroutine-storage']);
   b.version = 3;
   b.state.trackingEntries.push({ ...b.state.trackingEntries[0], id: 'bad-1', source: 'telepathy' });
   localStorage['zenroutine-storage'] = JSON.stringify(b);
   location.reload();
   ```
   (If the store has no tracking entries yet, load the sample data from Settings first.)
4. **Expected:** the app opens normally, and the quarantine notice appears. `JSON.parse(localStorage['zenroutine-quarantine'])` holds **exactly one** generation containing `bad-1` with its record verbatim, and `JSON.parse(localStorage['zenroutine-storage']).version` is back to `9` with `bad-1` gone from `state.trackingEntries`. The record is in the side-car before it left the blob.

**Observation B — the ordering guarantee (the actual fix).** This one cannot be seen without breaking something, which is why it lives in the negative controls above rather than here. On an unmodified build there is no input that reaches the closed window: that is the invariant, and it is what the new test file asserts. The honest summary is that the fix is verified by NC1b, not by a browser.

**What a reviewer should read instead of clicking:** `useAppStore.ts` `migrate` (the `hydratePersistedState` call and the comment above it), and the rewritten comment in `initializeAppStore` — those two are the whole change. The 611-line test file is the evidence for it.

## 6. Not done / unverified

**Physical-device smoke (#15) — still never run, and this is the hydration path.** What only a real device can settle, specifically:

- **A real torn AsyncStorage write.** `tests/setup.ts` mocks AsyncStorage as an in-memory `Map`, which is atomic per key. Every unparseable record and stale reference in the suite is hand-authored. The corruption shape this entire mechanism exists for — a half-written blob from a kill during `setItem` — has never been produced. On Android that's a SQLite row; on iOS a file plus a manifest entry; on web a `localStorage` slot. Three different tearing behaviours, none observed.
- **A real full disk / quota.** The tests simulate it by making `setItem` reject. I did verify by reading the source that on **web** a synchronous `QuotaExceededError` is converted to a **rejected promise** by `createPromise`'s try/catch, so `migrate`'s `await` does catch it — that much is measured, not assumed. The native paths are not.
- **Whether "awaited" means "durable".** The `await` in `migrate` guarantees the side-car write's promise resolved before the blob is rewritten. Whether resolution means committed to disk is a platform question (Android SQLite commit vs. iOS write + manifest rewrite vs. `localStorage` synchronous flush). Only a forced power cut mid-hydration settles it.
- **Whether both keys share a failure domain.** On web they share one `localStorage` origin quota, which means the fallback this fix relies on — "the record is still in the app blob" — may itself be unavailable exactly when the side-car write fails. Untested, and it is the most interesting thing a device or a quota-limited browser profile could tell us.
- **The migrate path itself has never been seen on a real store.** No device has ever held a ZenRoutine blob below v9. Every migrate-path test, including mine, is synthetic.

**Other limits, stated plainly:**
- **The closed window has no positive test, and cannot have one.** No input reaches it while the invariant holds — I checked every version gate in the file for one that is looser below CURRENT than at CURRENT, and `readTrackingPauses` is the only one, and it drops the key unread. So the closure is demonstrated by NC1b alone. If the orchestrator wants that permanently pinned, the only honest way is a fault-injection hook in production code, which I judged a worse trade than a recorded negative control.
- I did **not** add a runtime guard that throws if `merge` drops a record on the migrate path. It would have been untestable for the same reason, and an untestable new brick path is not an improvement. The structural change plus the two tested invariants cover it.
- No `src/core/engine/*`, no screens, no components, no `src/store/sampleData.ts`. I checked `_addSampleData` (`sampleData.ts`, #69) as the brief asked: it builds a whole state and `set`s it once, never touching `migrate`/`merge` or the quarantine sinks, so it bypasses nothing this change relies on.
- No dependency added; `package-lock.json` untouched.
- No push, no PR, no merge.


RogueKostas:
Landed by the orchestrator. CI green on 40620a8 (read before merging). Gate on the merged tree: typecheck exit 0, **725 tests / 44 files** under the host zone and TZ=UTC, build:web exported. No browser check: this lane changes no user-visible behaviour, and the lane said so plainly rather than inventing one. **NOT OBSERVED, and only a device can settle it (#15):** a real torn AsyncStorage write, a real quota failure on native, and whether an awaited write means a durable one.

